PR #18, opened 2013-10-03
changed setFrom function to convert newVal to a string when it is being ...

ChrisVoxland:
...used to set the value on html elements that don't take booleans. [Fixes #58086440]
@tobowers 
Not sure if you want me to add more tests, but if so please let me know what to test for and I'll write them.

### DIFF
--- a/test/variables.js
+++ b/test/variables.js
@@ -25,4 +25,16 @@ describe("Setting a single variable", function() {
     var template = generateTemplate({ singleVariable: "this is value" }, '<div><input class = "singleVariable-" /></div>')
     expect($(".singleVariable-").val()).to.be("this is value")
   })
+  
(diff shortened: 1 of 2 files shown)


test/variables.js
```diff
@@ -25,4 +25,16 @@ describe("Setting a single variable", function() {
     var template = generateTemplate({ singleVariable: "this is value" }, '<div><input class = "singleVariable-" /></div>')
     expect($(".singleVariable-").val()).to.be("this is value")
   })
+  
+  it("should set the value on select menus for true or false", function() {
+    var template = generateTemplate({ singleVariable: "false" },
```
tobowers, line 30: @ChrisVoxland  don't you want this to be a real false in order to test the new functionality

PR #19, opened 2013-10-07
have endDash trigger change events when it is updating the DOM.

tobowers:
When EndDash changes the DOM it doe _not_ actually fire a changeEvent (as you have to do that manually in JS, JQuery).  This change makes it so that when EndDash goes to change an element, it will fire the changeEvent.


jameselkins:
this looks good to me.

PR #20, opened 2013-10-08
Add a debugger directive to end-dash.

yaymukund:
Adding @tobowers as reviewer, cc @devmanhinton

We want to be able to run things in the context of the presenter once the data has loaded.

When do we start the debugger? I initially had it do a `setTimeout`, but that's kind of ugly and arbitrary. Now, I'm doing it when the model's loaded but it's possible that the presenter doesn't have a model. Maybe a `setTimeout` is fine given that we're in dev mode anyway...

### DIFF
--- a/lib/end-dash.js
+++ b/lib/end-dash.js
@@ -7,6 +7,7 @@ var Parser = require("./parser")
   , PartialReaction = require("./reactions/partial")
   , ViewReaction = require("./reactions/view")
   , ScopeReaction = require("./reactions/scope")
+  , DebuggerReaction = require('./reactions/debugger')
 
 module.exports = Parser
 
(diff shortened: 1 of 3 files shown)


tobowers:
what was confusing about the setTimeout is "where are you" when the debugger opens I think.  This seems much clearer to me if it's useful.  You would still have access to the stack I think and that'd be useful.

yaymukund:
So.. is it GTM? :]

I think the "presenter without a model" thing is a bigger issue. Is there a more appropriate event for this? `presenter.on('sync')` or something?

tobowers:
I think so except it's missing a test... this one might be really _hard_ to test though... thoughts?

yaymukund:
Oh, yeah, not sure, I will give it a stab.

yaymukund:
Since fireteam is busy, can you take a look at this @bglusman ? It's ready for review.

PR #22, opened 2013-10-11
Move templateRegistry/templateClassRegistry into new class, TemplateStore

yaymukund:
@tobowers, please review this sometime today or pass it on to someone else (bglusman?) if you're busy with higher priority stuff. (cc @devmanhinton) It's a decent sized branch, so I want to know of any red flags with how it's designed so I have time to fix it.

This branch decouples the `templateRegistry` and `templateClassRegistry` into its own class, `TemplateStore`. It should preserve current behavior.
1. Tests pass
2. I've QA'ed a bit, but this should be tested thoroughly on staging.
3. **TODO** I also need to add tests for the TemplateStore class.
4. **TODO** Add documentation. For example, `TemplateStore.load` should not actually parse the markup until we try to retrieve it. This isn't documented very well yet.

Once I do 2. and 3., we can merge this branch.

### DIFF
--- a/lib/template_store.js
+++ b/lib/template_store.js
@@ -1,0 +1,68 @@
+// This class is responsible for storing raw HTML templates when they're loaded
+// and then returning template objects (lazily).
+var _ = require('underscore'),
+    path = require('path'),
+    Parser = require('./parser'),
+    TemplateStore = {},
+    RAW_TEMPLATES = {},
+    TEMPLATES = {};
+
+var pathToName = function(templatePath) {
+  return path.normalize(templatePath);
+};
+
+var _parseTemplate = function(templatePath) {
+  var name = pathToName(templatePath);
+
+  if (!TemplateStore.isLoaded(name)) {
+    throw new Error('Could not find template: '+name);
+  }
+
+  var markup = RAW_TEMPLATES[name];
+
+  return (new Parser(markup, {
+    templateName: name,
+    templates: RAW_TEMPLATES
+  })).generate();
+};
+
+TemplateStore.isLoaded = function(templatePath) {
+  return !!RAW_TEMPLATES[templatePath];
+};
+
+TemplateStore.isParsed = function(templatePath) {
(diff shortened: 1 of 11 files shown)


devmanhinton:
Looks like a nice refactor to me very clean code, awesome decoupling :+1:

lib/template_store.js
```diff
@@ -0,0 +1,43 @@
+// This class is responsible for storing raw HTML templates when they're loaded
```
tobowers, line 1: there should be new tests for this class I think?

yaymukund, line 1: Yes, agreed. (I'm working on that after lunch.)

tobowers:
Forgive my ignorance... couple of questions...

What were we doing before to "store the templates" - that was just a require manager? 

If we keep all the HTML in memory will that be a problem for IE?  Or do we already do that anyway?

yaymukund:
@tobowers In terms of storage, I just moved `templateRegistry` to `RAW_TEMPLATES` and `templateClassRegistry` to `TEMPLATES`. That's where the markup/`Template` objects were stored, so we aren't storing any more than we were.

tobowers:
ah - then this looks good to me once there's the add'l tests

yaymukund:
@tobowers ready for code review/qa 

Okay, added some tests and made some refactors resulting from writing those tests. Noteworthy changes:
- Tests!
- `load` and `loadAndParse` will overwrite existing templates. We should test that this doesn't break stuff in prod.
- **TODO** in a subsequent branch/sprint, we should let users create instances of `TemplateStore`. Not high priority yet, but I might open up an issue.

yaymukund:
bump, could you review this sometime today please? We are building on top of this branch, so it would be helpful to merge this sooner rather than later.

I can ask someone else to review it if you're busy but I'm not sure who else is familiar with end-dash except bglusman (and I've already sent him one PR today... :)).

jameselkins:
this looks good to me.  Close & merge

PR #23, opened 2013-10-11
Open source steps

devmanhinton:
@tobowers @yaymukund  This is NOT ready to merge. Putting in a pull request for feedback.

Changes:
1. Add a Leader class -- end-dash.js that will be the top level abstraction users can see and use. When its loaded on a web page it will search the page for <script type="EndDash" name="whateverYouWant">...</script> and it will save whatever is between the script tags as a template internally with the script's name.
2. I added back end support for Javascript/Node as well as a less tested one for Ruby that will take a templates directory on the back-end and concat it into a single file (adding the name and surrounding script gas) that can then be served to the front-end
3. Added specs for all of this, although more specs would be nice

note: I also ran into what I believe is a jsdom bug (and that cost me some time) in that $("body").innerHTML = (endDash template) works but window.document.body.innerHTML = endDash template, does not. I pinged Zach on this since he put a pull request to fix this before.

Final goal is with this one can do:

On back end

aggregator = require(end-dash-aggregator)  (have not yet added the code to easily require it as a client)
aggregator.loadFiles("pathToTemplateDirectory", "locationYouWantTheServableFile")

//Load files into application.js and serve

Front End
End-Dash  = require("EndDash")
var end-dash = new EndDash() //All templates on page are loaded
var template = end-dash.createTemplate("NameofTemplate", modelToUse)
$("body").html(template.template)


xcoderzach:
> $("body").innerHTML = (endDash template) works but window.document.body.innerHTML = (endDash template)

Doing `$("body").innerHTML = (endDash template)` is just creating an innerHTML expando property on the jquery object.

Which totally can't be working :-S

devmanhinton:
hehe whoops, $("body").html(template.template) is working

lib/configured_parser.js
```diff
@@ -0,0 +1,22 @@
+var Parser = require("./parser")
```
yaymukund, line 1: Why do we need this class?

lib/server_side_dash/aggregator.rb
```diff
@@ -0,0 +1,61 @@
+class EndDashTemplateAggregator
```
yaymukund, line 1: I realize this code is a rough draft, but there are lots of problems with this— typos, camelCase, class variables, braces for multiline blocks, etc. 

Before you send this branch for review, please fix up those problems and document how this will actually be used in a Ruby project.

lib/compile.js
```diff
@@ -1,9 +1,9 @@
...
+var ConfiguredParser = require("./configured_parser")
   , fs = require("fs")
 
 var compile = module.exports = function(module, filename) {
   var content = fs.readFileSync(filename, 'utf8')
-  var Parsed = new EndDash(content, { templateName: filename.replace(/\.js\.ed(\.erb)?$/, "") })
+  var Parsed = new ConfiguredParser(content, { templateName: filename.replace(/\.js\.ed(\.erb)?$/, "") })
   module.exports = Parsed.generate()
 }
 if(require.extensions) {
```
yaymukund, line 9: Wasn't one of the main goals of this branch to get rid of the `require.extensions` load process?

lib/server_side_dash/aggregator.rb
devmanhinton, line 1: So.... it functions properly. Ie., directory of EndDash templates -> a single file with each template within script tags. I need to look into Rails asset pipeline.

lib/configured_parser.js
devmanhinton, line 1: We eventually can take it out. This is code in transition, this file was the old end-dash.js. This week I'll eliminate the configured parser and put this "configuring" power in the top level class, which actually makes more sense

lib/compile.js
devmanhinton, line 9: Nope! The main goal was to eliminate it from the template loading system. EndDash can be simply browserfied by either us before its released or else but the client. (In fact, when the scripts are aggregated this would be a logical place to browserify EndDash for the client and load it into the same EndDash file.

yaymukund, line 9: I don't understand your comment. Were you trying to eliminate `require.extensions` from the template loading system?

If so, how is "eliminating `require.extensions` from the template loading system" (your stated goal) different from "get rid of the `require.extensions` load process"?

devmanhinton, line 9: Sorry I misread your question. The goal is that you do not "have to" use the require system. require.extension is great. If someone has require in the browser yahelim! Power to em!

yaymukund, line 9: Hmm. I still see problems with messing with `require` like this. For one, [`require.extensions` is deprecated](http://nodejs.org/api/globals.html#globals_require_extensions). It's also a pretty invasive way to load templates.

In any case, I'm fine with not worrying about that right now. It's exactly the kind of tweak that can be handled in its own branch.

PR #24, opened 2013-10-14
Better grunt specs

yaymukund:
@bglusman, trying to send PRs to someone other than topper code reviews since he's busy...

cc @devmanhinton, feel free to chime in if you have any comments as well!

Right now, our tests only work because we load `helper.js` as a test file. Instead, let's explicitly require it. This also provides a nice place to put test helpers.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -10,22 +10,15 @@ module.exports = function(grunt) {
         reporter: 'spec'
       },
 
-      all: {src: ['test/**/*.js']}
+      all: {src: ['test/**/*.js', '!test/support/**/*.js']}
     },
 
     watch: {
       all: {
         files: ['test/**/*.js'],
-        tasks: ['simplemocha'],
-        options: {spawn: false}
+        tasks: ['simplemocha']
       }
     }
-  });
-
-  // Only run tests for the changed files. See:
-  // https://github.com/gruntjs/grunt-contrib-watch#compiling-files-as-needed
-  grunt.event.on('watch', function(action, filepath) {
-    grunt.config('simplemocha.all.src', filepath);
   });
 
   grunt.registerTask('default', ['simplemocha']);
--- a/docs/attribute_interpolation.markdown
+++ b/docs/attribute_interpolation.markdown
@@ -1,6 +1,6 @@
 [](root)
 ```coffeescript
-require("../test/helper")
+require("../test/support/helper")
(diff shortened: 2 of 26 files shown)


bglusman:
@yaymukund  looks pretty good, just to clarify, why are conditional_tag and conditional_attributes the only tests that required ./helper explicitly before?  is that the same file you moved down to ./support/helper for all the others to use?

Also, the two .markdown tests look to be the old AwesomeDocs tests, I think Topper replaced/converted all of those to mocha, are they still needed?  Seems pretty good aside from those couple questions.  Also curious about the addition of '!test/support/*_/_.js', assume that excludes files in test/support folder somehow, but curious how that fits into other changes etc since I'm not familiar with Gruntfile config stuff.

yaymukund:
@bglusman thanks for reviewing so soon, I really appreciate that.

> why are conditional_tag and conditional_attributes the only tests that required ./helper explicitly before?

Not totally sure. I actually think that zach just added it so he could run the test standalone, and then accidentally committed it. Just from looking at [the commit where he added it](https://github.com/Amicus/end-dash/commit/f1b2bcaa9e50b727c2f1d87001b8c054c9137a1b#diff-3b62c099ad9ddd54c460451b5112d46d), there's no obvious reason to require it only in that file.

> is that the same file you moved down to ./support/helper for all the others to use?

Yup

> Also, the two .markdown tests look to be the old AwesomeDocs tests, I think Topper replaced/converted all of those to mocha, are they still needed?

Good call. I'll remove those today in a separate branch? It's not really related to grunt or our tests.

> Also curious about the addition of '!test/support/*_/_.js', assume that excludes files in test/support folder somehow, but curious how that fits into other changes etc since I'm not familiar with Gruntfile config stuff.

Your instinct is correct— that just excludes files in `test/support`. I removed it because when we run tests, we should only run mocha test files. Under the hood, the grunt task basically [adds the files to the test run using `addFile`](https://github.com/visionmedia/mocha/wiki/Using-mocha-programmatically). It just so happens that calling `addFile` on our non-test files is harmless, but that could change.

yaymukund:
Gonna work on making this mergeable. Should be easy...

yaymukund:
@bglusman, I moved generateTemplate to its own file under test/support, and this is mergeable again now!

yaymukund:
bump @bglusman

PR #26, opened 2013-10-14
Combined branch

devmanhinton:
@yaymukund This merges our two branches. 

1) end-dash.js now exports an object that will be the API access to EndDash for any client.
2) Server side there is a "EndDashFileAggregator" which will turn a directory of templates into a single file
3) Template_store now has lazyLoadFromPage, which will store all <script type = "EndDash" name = "whatever">...</script> inside of template_store as raw HTML

All the specs were updated to reflect these changes. Mainly this was changing the tests in new/ and then util.js

I'll comment a few things I'd like feedback on


lib/template_store.js
```diff
@@ -55,14 +68,31 @@ TemplateStore.getTemplate = function(templatePath) {
...
+TemplateStore.lazyLoadFromPage = function (){
+  var that = this
+  _(getTemplatesOnPage()).each(function(template){
+    that.load(template.getAttribute('name'), normalizeTemplate(template))
+  })
+}
+
+// Lazily load for now, if we change this need to be careful of order
+// That is, first we need to load raw HTML and then parse otherwise it will
+// Error in the parser
```
devmanhinton, line 83: The parser looks up partials in RAW_TEMPLATES in the template_store passed via _state.
If a template with a partial is loaded into EndDash before the partial's template is
The partial will not be found and the parser will throw an error

Solution: Load the raw HTML of the templates first.
Then, when there is a call to TemplateStore.getTemplate it will parse the file 
It will find the all the partials because they have been loaded

test/new/cached_templates.js
```diff
@@ -0,0 +1,100 @@
...
+
+    it("EndDash should return a template with the model passed in bound to the DOM elements", function() {
+      var model = new Backbone.Model({name: "Devon"})
+      var template = EndDash.bindTemplate("helloWorld", model)
+      $("body").html(template.template)
+      expect($($('.name-')).text()).to.be("Devon")
+      model.set("name", "Brian")
+      expect($($('.name-')).text()).to.be("Brian")
+    })
+  })
```
devmanhinton, line 75: Below is code for a test that checks if templates beginning with words and not html elems work properly. They do not. Thus we can test for this condition and warp everything in an extra div or wrap the text in a span. Or simply document this to the user. I would opt for the later but want to hear your thoughts

lib/server_side_dash/aggregator.js
```diff
@@ -0,0 +1,65 @@
+var fs = require('fs')
```
devmanhinton, line 1: High level question: 

Order is:

Read all templates into EndDashFileAggregator object, into hash 'files'
Write all these to a new file

This could be:

Read one template file, write it. 

## +/-

Currently, can check to see if cached for future use but this is not really used yet. It also means much more in memory since all files are stored. The second way, there would be very little information in memory.

I'm leaning towards changing for the second way, but wanted to get your feedback

lib/configured_parser.js
```diff
@@ -0,0 +1,22 @@
+var Parser = require("./parser")
```
yaymukund: I thought we were eventually getting rid of this intermediate class [as per your comment in #23](https://github.com/Amicus/end-dash/pull/23#discussion_r6944727)?

devmanhinton: Get ride of the configured parser not parser.js, doing this now

lib/server_side_dash/aggregator.js
yaymukund, line 1: Code is worth a thousand words.. can you come up with a quick example of how to use this class? Shouldn't be very long.

lib/server_side_dash/aggregator.js
```diff
@@ -0,0 +1,65 @@
...
+  var fileNames = fs.readdirSync(this.templateDir)
+  if(fileNames.length == 0) {
+    throw new Error("no templates found in " + this.templateDir)
+  }
+  this.parseDirectory(fileNames)
+  this.writeFiles()
+}
+
+EndDashFileAgreggator.prototype.deleteOldFile = function() {
+  var path = this.aggregatedFilesDir + "/EndDashTemplates.js"
```
yaymukund, line 22: Filenames should be underscored rather than camelCased.

lib/reactions/partial.js
```diff
@@ -7,17 +7,22 @@ var Reaction = require("../reaction")
...
- 
+
 PartialReaction.preparse = function(el, state) {
   var src = el.attr("src")
     , file = path.resolve(state.currentDir(), src)
     , template = $(state.templates[file])
 
-  if(!template.length) {
-    throw new Error("could not find partial " + file + " in " +  _.last(state.pathStack))
+  if(!template.length) { // Nesting here is for backwards capability
```
yaymukund, line 16: What does this comment mean? o.O

lib/server_side_dash/aggregator.js
yaymukund, line 1: > Currently, can check to see if cached for future use but this is not really used yet.

You forgot some nouns. What is doing the checking? What is being cached? Where would it be cached?

devmanhinton, line 1: Sure. So 

``` javascript
Directory struct:
app.js
etc..
templates/
   home.js.ed
   call.js.ed
   poll/
     page1.js.ed
     page2.js.ed

var Aggregator = require('pathToAggregator') (will expose this in a easier way)
var aggregator = new Aggregator({teamplateDir: pathToTemples, aggregatedFilesDir: whereIwantFileOutput)
aggregator.loadFiles()

//Use whatever asset pipeline you want to serve to client side
```

Then -> a file EndDashTemplates.js w/
script type="EndDash" name="home"..../script
etc...
etc..
script type="EndDash" name="poll/page2".../script
^ no brackets since github will actually interpret as a script

This can be done in any .js file and then that file can be combined with others and served to the client side.

lib/server_side_dash/aggregator.js
devmanhinton, line 22: Do you mean the name of the variable "fileNames" and why? If so, that would be inconsistent. Everything else in that file is camelCased not underscored

yaymukund, line 22: no, I mean we should spit out `end_dash_templates.js` instead of `EndDashTemplates.js`.

lib/reactions/partial.js
devmanhinton, line 16: EndDash originally supported only templates and partial inclusion via the require system. Now it does not. Require will use the full pathname from root to file. In the new way (with script tags) the name will be relative to templates directory. Hence, this check. If you do not find it using the require name, us the script tag name. Wohoooo

lib/server_side_dash/aggregator.js
devmanhinton, line 22: Sure, I'm fine with that. Can you explain why that is better?

yaymukund, line 22: convention, consistency, predictability

tobowers, line 22: in general in almost all frameworks and languages -  file names are underscored (even if the language is usually camelCased).  Even JAVA does this JAVA I SAY!

lib/template_store.js
yaymukund, line 83: Lazy loading is a key feature for performance reasons alone. For example: We have like 100 templates but only use a handful on any given page. It just doesn't make sense to parse 100 templates if you're only going to use 5.

lib/server_side_dash/aggregator.js
yaymukund, line 22: *gasps*

lib/template_store.js
devmanhinton, line 83: Great! This was what I was thinking as well

lib/server_side_dash/aggregator.js
devmanhinton, line 22: hahah, cool! Will change that, thanks for the knowledge *brainSwells

devmanhinton:
Thanks for the feedback, all!

PR #27, opened 2013-10-15
Top level class

devmanhinton:
@yaymukund @tobowers 

I submit that this code is ready to merge. There are a lot of commits and not that much code added. This is simply my unfamiliarity with Javascript / time it took to merge Mukund's branch.

This version:

Has a top level EndDash class, that is the Queen/King of the castle (EndDash code)
It decouples the require system from EndDash, HTML in script tags of type EndDash can
be loaded into EndDash by calling EndDash.loadFromPage()

This functionality is spec-ed, see test/new but also all tests. The file util.js and the function generate template now uses the top level end-dash.js and is used in all tests.

Finally, I symlinked brief to this repo

screen shot 2013-10-15 at 11 15 42 am
^see the console

And tried: calling, editing a script, logging in logging out, being on the match friends page. And got / saw no errors.

Now, I'll be working on code meant for the server side, that will compile EndDash assets to a single file.


lib/template_store.js
```diff
@@ -55,14 +67,31 @@ TemplateStore.getTemplate = function(templatePath) {
...
   }
-
   return TEMPLATES[name];
 };
 
+//Lazy loading saves time since there are many templates per page and few used
+//Important, order of loading HTML and parsing matters. Those templates with
+//partials depend on their partials being loaded as raw HTML before they go
+//Through the parser
+TemplateStore.lazyLoadFromPage = function (){
```
yaymukund: This is defined at the wrong level. The `TemplateStore` should know nothing about the page. I would actually recommend putting this in a `page_helper.js` or something, far, far away from the rest of everything.

lib/template_store.js
```diff
@@ -55,14 +67,31 @@ TemplateStore.getTemplate = function(templatePath) {
...
+  var that = this
+  _(getTemplatesOnPage()).each(function(template){
+    that.load(template.getAttribute('name'), normalizeTemplate(template))
+  })
+}
+
+TemplateStore.clear = function() {
```
yaymukund, line 61: same as above. this is only used in tests, so it should be defined in `test/util.js`.

lib/template.js
```diff
@@ -19,7 +19,7 @@ function Template(model, opts) {
   if(!this.markup) {
     throw new Error("Created template without markup")
   }
-  this.template = this.markup.clone()
+  this.el = this.template = this.markup.clone()
```
yaymukund: same as above

lib/end-dash.js
```diff
@@ -11,9 +11,47 @@ var _ = require('underscore'),
...
+EndDash.clearAndReload = function() {
+  TemplateStore.clear()
+  this.loadPage()
+}
+
+EndDash.loadFromPage = function() {
+  TemplateStore.lazyLoadFromPage()
+}
+
+EndDash.isTemplateLoaded = function(templatePath) { //Here for testing
```
yaymukund: Same as above.

lib/end-dash.js
```diff
@@ -11,9 +11,47 @@ var _ = require('underscore'),
...
+EndDash.boundTemplate = function (templateName, model) {
+  var TemplateClass = TemplateStore.getTemplate(templateName)
+  return new TemplateClass(model)
+}
+
+EndDash.registerTemplate = function(templatePath, markup) {
+  return TemplateStore.load(templatePath, markup)
+}
+
+EndDash.getTemplateClass = EndDash.getTemplate = function(templatePath){
```
yaymukund: If we're aliasing methods, we should have a good reason and it should be commented.

lib/end-dash.js
```diff
@@ -11,9 +11,47 @@ var _ = require('underscore'),
...
+  return TemplateStore.getTemplate(templatePath)
+}
+
+EndDash.clearAndReload = function() {
```
yaymukund: We shouldn't define methods in the main object if htey're only going to be used in tests. You should move this to `test/util.js`.

lib/end-dash.js
```diff
@@ -11,9 +11,47 @@ var _ = require('underscore'),
...
+  Parser.registerReaction(ModelReaction);
+  Parser.registerReaction(AttributeReaction);
+  Parser.registerReaction(VariableReaction);
+  Parser.registerReaction(ConditionalReaction);
+  Parser.registerReaction(ViewReaction);
+  Parser.registerReaction(DebuggerReaction);
+
+EndDash = {}
+
+EndDash.boundTemplate = function (templateName, model) {
```
yaymukund: The whole `getTemplate` versus `getTemplateClass` thing is very confusing. How about:

``` javascript
// Rename TemplateStore.getTemplate to TemplateStore.getTemplateClass
TemplateStore.getTemplateClass(templateName);

// Fetch a template instance and optionally bind it to a model.
EndDash.getTemplate(templateName, model=null);

// Bind a model to a template.
template.bind(model);
```

lib/end-dash.js
devmanhinton: Brief relies on getTemplate to be defined but getTemplateClass makes more sense. This is a framework so there should be no reference to brief thus no comment. Once we change the function call in brief we can remove this. For not, to be safe, I think this is best as it still allows EndDash to work with brief as is, and does not add comments specific to our stack

lib/template.js
devmanhinton: Disagree with this. This is simply ease of naming. Not sure where else to abstract that would make sense. The template and and markup are tightly coupled. Reference the markup by the template name is common in brief so we still need that but el makes a lot more sense.

lib/template_store.js
devmanhinton: I like that. Good call Mukund! Will add that after lunch and push it up

lib/end-dash.js
yaymukund: You don't have to mention brief... just say "bar is deprecated, please use foo."

lib/template_store.js
devmanhinton, line 61: Mmmm Do we really want no clearing functionality, It seems natural to allow a user to clear the templates. What harm do you think there is to exposing such functionality?

lib/end-dash.js
devmanhinton: Same as above from me too!

lib/end-dash.js
devmanhinton: Agreed! Again see above two comments.

lib/end-dash.js
devmanhinton: Yeah, we can do this.

lib/template.js
yaymukund: I meant same as [this comment from above](https://github.com/Amicus/end-dash/pull/27/files#r6975248)

What do you mean by "ease of naming"? I'm not sure what you think I'm asking you to decouple...

yaymukund: I reread this and tried to understand. If you think el is more semantic (I would agree!), can you just add a short comment like `// template is deprecated, but it's still referenced in some outside systems`.

devmanhinton: I think we are not talking about the same thing. 

this.el = this.template = this.markup.clone(), no?

devmanhinton: Yep, know we are the same page, we have a different way of communicating it seems. Will add the comment

lib/end-dash.js
devmanhinton: This will be in the next pull request, already in code locally

devmanhinton:
@yaymukund So I went through and changed all of your comment save one, I left EndDash.bindTemplate("templateName", model). I like

``` javascript
// Rename TemplateStore.getTemplate to TemplateStore.getTemplateClass
TemplateStore.getTemplateClass(templateName);

// Fetch a template instance and optionally bind it to a model.
EndDash.getTemplate(templateName, model=null);

// Bind a model to a template.
template.bind(model);
```

But getTemplate must return a TemplateClass object to work brief.

We can simply change the name and have this extra functionality, what do you think: leave as EndDash.bindTemplate for now or else add a new function like:

``` javascript
// Fetch a template instance and optionally bind it to a model.
EndDash.returnTemplate(templateName, model=null);

// Bind a model to a template.
template.bind(model);
```

yaymukund:
Are tests passing? I checked out c52c8ce, ran tests, and got:

```
  24 passing (318ms)
  1 failing

  1) With a set of templates loaded into EndDash Binding a template on the page "before each" hook:
     TypeError: Object [object Object] has no method 'add'
      at Context.<anonymous> (/Users/mukund/src/end-dash/test/end-dash.js:43:34)
      at Hook.Runnable.run (/Users/mukund/src/end-dash/node_modules/mocha/lib/runnable.js:211:32)
      at next (/Users/mukund/src/end-dash/node_modules/mocha/lib/runner.js:249:10)
      at Object._onImmediate (/Users/mukund/src/end-dash/node_modules/mocha/lib/runner.js:261:5)
      at processImmediate [as _immediateCallback] (timers.js:330:15)



Warning: Task "simplemocha:all" failed. Use --force to continue.

Aborted due to warnings.
```

devmanhinton:
@yaymukund Yes, they are all passing locally. Must be something locally or a merge you did. (if you did any)
screen shot 2013-10-15 at 6 24 27 pm

yaymukund:
brian got the same result

```
Brian G.    running now, slow on npm installing
mukund: 1 fail
View paste 
1) With a set of templates loaded into EndDash Binding a template on the page "before each" hook:
     TypeError: Object [object Object] has no method 'add'
```

devmanhinton:
What version of backbone do you have?

btws line is https://github.com/Amicus/end-dash/blob/top_level_class/test/end-dash.js#L43

Its adding a model to a backbone collection. Exact same code locally for me and it runs so I'm guessing it most be something to do with setup

yaymukund:
Have you tried rm rfing your `node_modules` and retrying already?

devmanhinton:
@yaymukund  Yep that errored, good call. I'm flipping back and forth between this and working with Topper. So when I can I will fix (a small thing I'm guessing) and repush it.

yaymukund:
> But getTemplate must return a TemplateClass object to work brief.

Are you saying that brief calls `getTemplate` and expects it to return a `Template` class?

devmanhinton:
https://github.com/Amicus/brief/blob/master/app/assets/javascripts/new/routes/organization_features.js.module.coffee#L11

^ and two other routers.

test/util.js
```diff
@@ -1,16 +1,29 @@
...
+    PageHelper = require('../lib/page_helper'),
     testTemplateCount = 0;
 
+  require('./helper')
+
+EndDash.isTemplateLoaded = function(templatePath) {
+  return TemplateStore.isLoaded(templatePath)
+}
+
+EndDash.clearAndReload = function() {
```
yaymukund, line 12: I thought we were going to get rid of this and `TemplateStore.clear()` based on our discussion in chat?

yaymukund:
Cool, makes sense. I think you made the right choice, but we should eventually just deprecate `bindTemplate` in favor of `getTemplate` but that can wait until a subsequent branch.

yaymukund:
See #32 instead

PR #28, opened 2013-10-15
Make TemplateStore a factory rather than a singleton class.

yaymukund:
Less global state! Just store the `templateStore` on the `EndDash` object. The test/util.js needs to be refactored after this, but that's lower priority.

@devmanhinton, @tobowers for review

### DIFF
--- a/lib/template_store.js
+++ b/lib/template_store.js
@@ -3,44 +3,42 @@
 var _ = require('underscore'),
     path = require('path'),
     Parser = require('./parser'),
-    TemplateStore = {},
-    RAW_TEMPLATES = {},
-    TEMPLATES = {};
+    TemplateStore = {};
 
 var pathToName = function(templatePath) {
   return path.normalize(templatePath);
 };
 
-var _parseTemplate = function(templatePath) {
+var _parseTemplate = function(templateStore, templatePath) {
   var name = pathToName(templatePath);
 
-  if (!TemplateStore.isLoaded(name)) {
+  if (!templateStore.isLoaded(name)) {
     throw new Error('Could not find template: '+name);
   }
 
-  var markup = RAW_TEMPLATES[name];
+  var markup = templateStore._RAW_TEMPLATES[name];
 
   return (new Parser(markup, {
     templateName: name,
-    templates: RAW_TEMPLATES
+    templates: templateStore._RAW_TEMPLATES
   })).generate();
 };
 
 TemplateStore.isLoaded = function(templatePath) {
-  return !!RAW_TEMPLATES[templatePath];
+  return !!this._RAW_TEMPLATES[templatePath];
 };
 
 TemplateStore.isParsed = function(templatePath) {
-  return !!TEMPLATES[templatePath];
+  return !!this._TEMPLATES[templatePath];
 };
 
 TemplateStore.load = function(templatePath, markup) {
   var name = pathToName(templatePath);
 
-  RAW_TEMPLATES[name] = markup;
+  this._RAW_TEMPLATES[name] = markup;
 
   if (this.isParsed(name)) {
-    delete TEMPLATES[name];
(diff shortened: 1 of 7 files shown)


lib/template_store.js
```diff
@@ -3,44 +3,42 @@
 var _ = require('underscore'),
     path = require('path'),
     Parser = require('./parser'),
-    TemplateStore = {},
-    RAW_TEMPLATES = {},
-    TEMPLATES = {};
+    TemplateStore = {};
```
devmanhinton, line 7: What is the advantage of defining all the functions on TemplateStore object and then exporting them in anonymous model. This functionality can clearly be achieved with prototypical inheritance or simply having all top-level scope function objects stored in variables (var isLoaded instead of TemplateStore.isLoaded for example). The argument I would see for this is that it allows you to separate core functionality of a templatestore (such as isLoaded) from helpers such as _parseTemplates. Would be interested to hear your rationale?

yaymukund, line 7: I'm confused, can you sketch out your design and give an example of how it would be better?

Right now, `_parseTemplates` is already separated from the public interface since it's not exported as a method on TemplateStore.

yaymukund, line 7: (BTW I'm still getting comfortable with prototypal inheritance. The more I read, the more I realize that it's a fine pattern and maybe we should use it but I don't see any strong advantages to convert this right this minute.)

devmanhinton, line 7: O yeah, definitely. I think this module is well designed. It is an interesting hybrid design style. Its similar to prototypical inheritance in that the main functions are defined on a single object that share the name of the module (ie., TemplateStore). 

This module would function the same way (no longer a singleton class, if we added .prototype between every function defined on Template store and then added

``` javascript
TemplateStore = function() {
  this._RAW_TEMPLATES = {}
  this._TEMPLATES = {}
}
module.exports = TemplateStore
```

However, we then export all the functionality we want in object literal.

``` javascript
exports.create = function() {
  return {
    _RAW_TEMPLATES: {},
    _TEMPLATES: {},
    isLoaded: TemplateStore.isLoaded,
    isParsed: TemplateStore.isParsed,
    load: TemplateStore.load,
    loadAndParse: TemplateStore.loadAndParse,
    getTemplate: TemplateStore.getTemplate
  };
};
```

Also, small style query, you end function definitions with ";", should I adopt that habit as well?

devmanhinton:
@yaymukund Can we get rid of this request? 

You already added and merged in a separate PR, correct?

devmanhinton:
oops, I'm sorry. This is different then what is currently master. Sorry about that. I'll take a look at this when I get a chance.

devmanhinton:
@yaymukund @bglusman This code is good. 

I was hoping to meet and discuss singleton vs classes & objects before merging. This is a pivotal ticket https://www.pivotaltracker.com/story/show/59594344

devmanhinton:
It'd be great if you could respond with a time today or tomorrow for this discussion

yaymukund:
@devmanhinton, Why do we need to talk about implementation?

That task is slightly misleading. We need to make a decision on singleton v. non singleton. I think you should do some research, send a PR that refactors one object, and we can talk there.

If you still think it's important to meet, we can talk for 5-10m at 3:30pm?

devmanhinton:
@yaymukund 

Because two of our team goals are to make a decision on

1) Singleton vs non-singleton
2) Prototypical vs non-prototypical inheritance 

If we do not meet to discuss/decide how are we deciding these questions?

Unless discussed we should stick to the tickets created in pivotal. If anyone disagrees with the tickets
(Created Friday to each match a single team goal) he or should should bring it up
in campfire or via some other medium.

bglusman:
Closing for now in favor of https://github.com/Amicus/end-dash/pull/47

PR #29, opened 2013-10-17
Template directory helper js

devmanhinton:
@yaymukund I reworked this to make is async and use a bunch of currying. It is my first stab at currying or really async js so feedback is appreciated. 

How it works:
It recursively counts endDash templates in the inputted directory. This number is used to determine when to hit the callback. It then iterates through every file and, if it ends in .js.ed or .html, it writes the contents of the file to a end_dash_templates.js file.

### DIFF
--- a/lib/formate_templates.js
+++ b/lib/formate_templates.js
@@ -1,0 +1,74 @@
+var fs = require('fs')
+  , path = require('path')
+  , _ = require('underscore')
+  , outputPath = __dirname
+  , filesToParse = 0
+  , callback
+
+var compressFiles = function(templatesLocation, outputPath, callback) {
+  if (arguments.length != 3) {
+    throw new Error("All three arugments are required")
+  }
+  var templatesLocation = addLeadingSlash(templatesLocation)
+    , ws = fs.createWriteStream(addLeadingSlash(outputPath) + "end_dash_templates.js", "w")
+    , callback = callback
+  traverseFilesSync(templatesLocation, null, function(){filesToParse++})
+  traverseFilesSync(templatesLocation, null, readAndWriteFile_curry(ws, callback))
+}
+
+var traverseFilesSync = function(dir, dirCallback, fileCallback) {
+  var files = fs.readdirSync(dir)
+  _.each(files, function(fileName) {
+    if(fs.statSync(dir + fileName).isFile()){
+        if (fileName.match(/.*.js.ed\s*$/) || fileName.match(/.*.html\s*$/) ) {
+          fileCallback && fileCallback(dir, fileName)
+        }
+    } else {
+      dirCallback && dirCallback(dir, fileName)
+      traverseFilesSync(dir + fileName + "/", dirCallback, fileCallback)
+    }
+  })
+}
+
+var readAndWriteFile_curry = function(ws, callback) {
+  return function(dir, fileName) {
+    fs.readFile(dir + fileName, "utf-8", writeCallback_curry(ws, dir + fileName, callback))
+  }
+}
+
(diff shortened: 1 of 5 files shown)


devmanhinton:
@tobowers Any thoughts on this code?

It would be used to load server_side templates to a single file.

lib/formate_templates.js
```diff
@@ -0,0 +1,74 @@
...
+  })
+}
+
+var readAndWriteFile_curry = function(ws, callback) {
+  return function(dir, fileName) {
+    fs.readFile(dir + fileName, "utf-8", writeCallback_curry(ws, dir + fileName, callback))
+  }
+}
+
+var writeCallback_curry = function(ws, path, callback) {
```
tobowers, line 39: mixing underscores is a little weird.

test/formate_template.js
```diff
@@ -0,0 +1,42 @@
+var formate_templates = require('../lib/formate_templates')
```
tobowers, line 1: typo?  "format" ?

yaymukund:
@devmanhinton, sorry, will chime in w/ some thoughts tonight. This is lower priority but no PR should be stuck in limbo for more than a day, imo. Please nag me first in the future :] :tanabata_tree:

lib/formate_templates.js
```diff
@@ -0,0 +1,74 @@
...
+  , filesToParse = 0
+  , callback
+
+var compressFiles = function(templatesLocation, outputPath, callback) {
+  if (arguments.length != 3) {
+    throw new Error("All three arugments are required")
+  }
+  var templatesLocation = addLeadingSlash(templatesLocation)
+    , ws = fs.createWriteStream(addLeadingSlash(outputPath) + "end_dash_templates.js", "w")
+    , callback = callback
```
tobowers, line 14: @yaymukund what's your thought on using some sort of promises library to cut down on the passing of callbacks?

lib/formate_templates.js
devmanhinton, line 39: Should I change it? I did that so to purposefully set these apart to indicate they were curry functions

test/formate_template.js
devmanhinton, line 1: Yes, typo

lib/formate_templates.js
tobowers, line 39: readAndWriteFileCurry still separates it I think.  Still not convinced that we don't want to use a library here to cleanup the syntax.

yaymukund:
Hmm, I've read through this and I think we're doing far too much on our own. I think the easiest/cleanest solution for now is to use [grunt](http://gruntjs.com/) and the [grunt-contrib-concat plugin](https://github.com/gruntjs/grunt-contrib-concat). There's [an example on how to process files before concatenating them](https://github.com/gruntjs/grunt-contrib-concat#custom-process-function).

You'll get for free:
- easily select files to concatenate using [the powerful minimatch library](https://github.com/isaacs/minimatch).
- far less code to write... nothing to test.
- easily combine it with other tasks such as watching for changes and restarting the server.

It might eventually make sense to remove the grunt dependency and build a real CLI tool or an API, but I'm inclined to take the easiest approach until we finalize this format. The grunt approach would literally be something like:

``` javascript
grunt.initConfig({
  concat: {
    dist: {
      options: {
        process: function(src, filepath) {
          return '<script type="text/enddash" name="'+filepath+'">'+src+'</script>';
        },
      },
      files: {
        'dist/built.js': ['src/project.js'],
      },
    },
  },
});
```

:sandal:

yaymukund:
@devmanhinton, hey, I'm closing this up since it hasn't been touched in awhile. Feel free to reopen if you want.

PR #30, opened 2013-10-17
Top level class noncompat

devmanhinton:
@yaymukund CR, here is the top_level_class branch but with EndDash.getTemplate returning an instance. This can be called with a model or without one. 

With a model--it binds the model to the template and returns it. 
Without a model--it returns an unbound template


lib/end-dash.js
```diff
@@ -11,9 +12,44 @@ var _ = require('underscore'),
...
+
+EndDash.getTemplateClass = function(templatePath) {
+  return TemplateStore.getTemplate(templatePath)
+}
+
+EndDash.getTemplate = function(templatePath, model) {
+  var TemplateClass = this.getTemplateClass(templatePath)
+  return new TemplateClass(model)
+}
+
```
devmanhinton, line 44: Above two functions changed

lib/template.js
```diff
@@ -19,25 +18,35 @@ function Template(model, opts) {
...
+  }
+}
+
+Template.prototype._bind = function(model) {
+  this.stack.push(model)
   this.traverse(this.structure, this.stack, function(el, reaction, stack, next) {
     reaction.start(el, stack, next)
   })
+  this.bound = true
 }
```
devmanhinton, line 51: Change .bind -> ._bind

lib/template.js
```diff
@@ -19,25 +18,35 @@ function Template(model, opts) {
   if(!this.markup) {
     throw new Error("Created template without markup")
   }
-  this.template = this.markup.clone()
-  this.bind()
+  this.el = this.template = this.markup.clone() // template is deprecated but used in some outside systems
+  if (model) {
+    this._bind(model)
+  }
 }
```
devmanhinton, line 25: Check for model before binding

devmanhinton:
Added some comments to help isolate the new changes from top_level_class. Main changes are in template.js

lib/template.js
```diff
@@ -19,25 +18,35 @@ function Template(model, opts) {
...
+  })
 }
 
 Template.prototype.getElement = function(selector) {
   return findDescendantsAndSelf(this.template, selector)
 }
 
-Template.prototype.bind = function() {
+Template.prototype.bind = function(model) {
+  if(model && this.bound == false) {
```
yaymukund, line 39: Why move `_bind` into its own function?

Also, you should check `!this.bound` and not `this.bound == false`.

Never compare booleans using `==` or `===`...

test/util.js
```diff
@@ -1,26 +1,49 @@
...
   } else {
     templatePath = generateTestTemplatePath();
-    Template = TemplateStore.loadAndParse(templatePath, markupOrPath);
+    EndDash.registerTemplate(templatePath, markupOrPath);
   }
 
-  var template = new Template(model, {
-    templateName: templatePath+'.html'
-  });
+  if (whichTemplateLoad == 0) { // Alternate how template is created
```
yaymukund, line 28: We should avoid "clever" code.

If one of the template creation methods has a bug, for example, this would trigger a failure in every other test. It wouldn't matter _what_ the tests were, and changing the order of the test would cause different tests to fail.

If you want to test that both template creation processes work (a noble goal), write actual specs for them.

lib/end-dash.js
```diff
@@ -11,9 +12,44 @@ var _ = require('underscore'),
...
+EndDash.bindTemplate = function(templatePath, model){
```
yaymukund, line 45: Why is this still here? Doesn't `getTemplate` deprecate this?

test/end-dash.js
```diff
@@ -0,0 +1,181 @@
...
+_ = require('underscore')
+
+var partials = fs.readFileSync(__dirname + "/support/partials.html", 'utf8')
+var embedded_partial = fs.readFileSync(__dirname + "/support/embedded_partial.html", 'utf8')
+var list_item = fs.readFileSync(__dirname + "/support/list_item.html", 'utf8')
+var scopes = fs.readFileSync(__dirname + "/support/scopes.html", 'utf8')
+
+describe("With a set of templates loaded into EndDash", function(){
+  beforeEach(function(){
+    EndDash.registerTemplate("partials", partials)
```
yaymukund, line 17: `registerTemplate` is deprecated. You should not use it.

test/end-dash.js
```diff
@@ -0,0 +1,181 @@
...
+  , Backbone = require("backbone")
+  , EndDash = require('../lib/end-dash')
+  , fs = require('fs')
+  , util = require("./util")
+
+_ = require('underscore')
+
+var partials = fs.readFileSync(__dirname + "/support/partials.html", 'utf8')
+var embedded_partial = fs.readFileSync(__dirname + "/support/embedded_partial.html", 'utf8')
+var list_item = fs.readFileSync(__dirname + "/support/list_item.html", 'utf8')
```
yaymukund, line 12: `camelCase`

test/util.js
```diff
@@ -1,26 +1,49 @@
-var TemplateStore = require('../lib/template_store'),
+var EndDash = require('../lib/end-dash'),
+    TemplateStore = require('../lib/template_store'),
+    PageHelper = require('../lib/page_helper'),
+    whichTemplateLoad = 0
     testTemplateCount = 0;
 
+  require('./helper')
+
+EndDash.isTemplateLoaded = function(templatePath) {
```
yaymukund, line 9: Err, you don't need to alias this. Just call `TemplateStore.isLoaded` directly.

test/util.js
devmanhinton, line 28: Yeah, was debating that / interested to hear your take. I thought it was a cool way to much better test the functionality but are right that it is unclear from a glance what is occurring. Maybe add another "generateTemplate" function that would use the second one of creating templates

lib/end-dash.js
devmanhinton, line 45: No, they have different purposes, without getTemplateClass you have no one to get the TemplateClass. Do we not want to expose this to the user? I may be biased since this is how we did it in brief.

test/end-dash.js
devmanhinton, line 17: ?? Since when is registerTemplate deprecated??

test/util.js
devmanhinton, line 9: Yeh, either way works

lib/end-dash.js
yaymukund, line 45: This has nothing to do with `getTemplateClass`... I'm asking you if we can remove `bindTemplate`.

test/util.js
devmanhinton, line 9: Actually no, not in this branch unless you require('TemplateStore') in the testFile. This seems bad? We want tests to be as similar to usual behavior as possible and we never want the user to directly interact with the TemplateStore. I suppose since it is a test that is fine...? 

This is also all going to change soon once we merge in your "getting rid of singletons'

test/util.js
yaymukund, line 28: > Maybe add another "generateTemplate" function that would use the second one of creating templates

Why?

lib/end-dash.js
devmanhinton, line 45: Ahhh, yep. You are right. Thought you were referencing getTemplateClass for some reason. For some reason, your comment applying to the line right above it didn't register in my brain. It should be removed and then removed from the specs

test/end-dash.js
devmanhinton, line 12: true

test/util.js
yaymukund, line 9: Yes, it's fine for tests.

test/util.js
devmanhinton, line 28: Well getTemplateClass and getTemplate run very similar code so its not a big deal but right now 90% or higher tests use EndDash.getTemplateClass via support/util.js generateTemplate. Why not switch it up every once in a while in order to spread out the functionality tests hit.

test/end-dash.js
yaymukund, line 17: Oops, you're totally right. My bad.

test/util.js
yaymukund, line 28: I already addressed this:

> If one of the template creation methods has a bug, for example, this would trigger a failure in every other test. It wouldn't matter what the tests were, and changing the order of the test would cause different tests to fail.
> 
> If you want to test that both template creation processes work (a noble goal), write actual specs for them.

lib/template.js
devmanhinton, line 39: This is more a result of the process then how it needs to be. At first there was a lot more functionality in bind so I moved the core into _.bind. But as I refactored and continuously got specs to pass bind got whittled further and further down. At this point _.bind and .bind could be combined. I did not have time to do this before my call.

yaymukund, line 39: > At this point _.bind and .bind could be combined.

K.

test/util.js
devmanhinton, line 28: umm That is what adding the second generate function would accomplish..

Sure writing more specs would work but the specs written already test every part of EndDash functionality. What not utilize the already written specs to greater advantage rather then writing more specs?

yaymukund, line 28: Because specs should only test one thing. If a spec called "collections with more than one item render correctly" fails, I want it to be because there's a bug with rendering collections with more than one item and not because of some low-level template creating bug.

devmanhinton, line 28: Definitely agree, and that is why in the end, alternating between the two is not a good idea. I can definitely see your point "If two specs fail that are testing the same thing I want to isolate if it is the test or the underlying template generation is the problem for both (ideally the former)." I think the way I described has merits as well, but for the sake of expediency let us move forward and just write moar specs!

test/end-dash.js
devmanhinton, line 17: No worries at all

yaymukund:
see #32 instead

PR #31, opened 2013-10-17
Clean up tests and code.

yaymukund:
Hi @devmanhinton, 

Did a big refactor of the top_level_class_noncompat branch based on our discussion in the last CR. I simplified the tests a ton to make them more readable and generally reduce their complexity.

This is the diff between my branch and yours. You can also [view the diff against master](https://github.com/Amicus/end-dash/compare/end-dash-public-interface).

Please let me know if you think something should be changed. Otherwise, we can work on squashing and merging this into master tomorrow.


lib/end-dash.js
```diff
@@ -1,54 +1,38 @@
...
+  return new Template(model);
 }
 
-// This is a hack to make requiring a file that ends in .js.ed
-// to return a EndDash compiled version of that file
+// This is a hack to make requiring a file that ends in .js.ed to return a
+// EndDash compiled version of that file. We need to expose it publicly,
+// because we have code that relies on it, but it's deprecated and not
+// officially a part of the public API.
 EndDash.compile = function(module, filename) {
```
devmanhinton, line 36: hmmm. Is this true? Do we not want to support EndDash using require? If they have require in the browser (via browserify or something along those lines, I'm not to familiar with all this)

lib/template.js
```diff
@@ -5,23 +5,23 @@ var Backbone               = require("backbone")
...
+  var stack;
 
-  opts = opts || {}
-  this.stack = opts.stack || []
-  this.bound = false
+  opts = opts || {};
+  model = model || {};
 
+  this.stack = opts.stack || [];
+  this.stack.push(model);
```
devmanhinton, line 15: Not sure about this. We are allowing for 

``` javascript
var template = EndDash.getTemplate("foo")
template.bind(model)
```

Does this still work? From what I can tell even if no model is given, this will bind the template instance to an empty object. Even if rebinding is okay, you are pushing an empty object on the stack that is not popped. This seems okay for most cases but what if partials have a stack passed in (which they might), adding the extra empty model may break things. This needs to be tested.

devmanhinton:
@yaymukund Seems good overall, nice refactor of specs! There was one minor comment and then a more major one. See above for details

lib/end-dash.js
yaymukund, line 36: Correct! Not worth the trouble to maintain.

lib/template.js
yaymukund, line 15: The old code was basically doing:

``` javascript
Template.prototype._bind = function(model) {
  this.stack.push(model)
     reaction.start(el, stack, next)
   })
   this.bound = true
};
```

Does this work? Since we only ever push to the stack, old models stick around. Could that be an issue?

yaymukund:
gonna close this and work on fixing `bind(model)` as we'd discussed in person.

PR #32, opened 2013-10-18
Clean up enddash public API.

yaymukund:
Hi @devmanhinton, @tobowers 

We reworked the EndDash public API and made it ready to be used by a developer. We need to add documentation and an example application, which we'll do Today and Monday (a day or two behind because we've both been on emergency stuff). A quick summary:

``` javascript
// Load templates from <script type="text/enddash"></script> on
// the page.
EndDash.loadTemplatesFromPage = PageHelper.loadFromPage;

// Add a template to the store. Do not parse it.
EndDash.registerTemplate = TemplateStore.load;

// Retrieve a template class. Brief uses this. We should try to
// deprecate this and push people towards `getTemplate`
EndDash.getTemplateClass = TemplateStore.getTemplateClass;

// Retrieve a template instance, optionally bound to a model.
EndDash.getTemplate = function(templatePath, model) {
  var Template = this.getTemplateClass(templatePath);
  return new Template(model);
}
```

If this is good, we should merge it and make it work with brief.

### DIFF
--- a/lib/page_helper.js
+++ b/lib/page_helper.js
@@ -1,0 +1,20 @@
+var _ = require('underscore')
+  , templateStore = require('./template_store');
+
+// Only load HTML do not parse until template is requested
+// If this is changed be aware, order matters. If templates with
+// partials are parsed before their partials raw HTML are parser.js will error
+exports.loadFromPage = function () {
+  $('script[type="text/enddash"]').each(function() {
+    var $el = $(this),
+        // Need to trim whitespace or else jQuery will complain.
+        markup = $.trim($el.html()),
+        name = $el.attr('name');
+
+    if (!name) {
+      throw new Error("Script tags of type text/enddash must have a 'name' attribute");
+    }
+
+    templateStore.load($el.attr('name'), markup);
+  });
+};
--- a/test/end-dash.js
+++ b/test/end-dash.js
@@ -1,0 +1,92 @@
+require('./support/helper');
+
+var expect = require('expect.js'),
+    Backbone = require('backbone'),
+    EndDash = require('../lib/end-dash'),
+    TemplateStore = require('../lib/template_store'),
+    Template = require('../lib/template'),
+    _ = require('underscore');
+
+describe('EndDash', function(){
+  describe('.registerTemplate', function() {
+    it('loads the template to the store', function(){
+      EndDash.registerTemplate('partials', '<div></div>');
+      expect(TemplateStore.isLoaded('partials')).to.be(true);
+    });
+  });
+
+  describe('.getTemplate', function() {
+    it('retrieves loaded templates', function() {
+      EndDash.registerTemplate('dog', '<div name="unbound-"></div>');
+      var template = EndDash.getTemplate('dog');
+      expect(template instanceof Template).to.be(true);
+      expect(template.el.text()).to.be('');
+    });
+
+    it('binds values from the model to the template', function() {
+      EndDash.registerTemplate('cat', '<div class="catName-"></div>');
+
+      var model = new Backbone.Model({catName: 'Alabama'}),
+          template = EndDash.getTemplate('cat', model);
+
+      expect(template.el.text()).to.be('Alabama');
+    });
+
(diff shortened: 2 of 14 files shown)


tobowers:
sorry... let me just type this back to you so I make sure I understand...

let's say I have a template:

```
sweet_template.js.ed = '<div class="sweet-"></div>'
```

where sweet_template.js.ed is the file name.

I could put it on the page like so:

```
<script id="sweet_template.js.ed" type="text/enddash">...</script>
```

I would then load it with EndDash.loadTemplatesFromPage("sweet_template.js.ed")?

In general then when I wanted to get a template I would now do this awesome thing:

EndDash.getTemplate("sweet_template", new Backbone.Model()) which would give me this sick as shit cool ass sweet template bound to that new model?

yaymukund:
**EDIT** Small correction: `loadTemplatesFromPage` doesn't take an argument. You just call it and it will scour the page for all `text/enddash` script tags.

Yes, that's the simplest way to load templates (for docs, jsbin, small apps, messing around...).

You can also call `EndDash.registerTemplate('sweet_template.js.ed', '<div class="foo-"></div>');` from anywhere in your app. This lets people build tools on top of end-dash.

We're working on a Rails gem that will package all your templates into a single `templates.js.ed` which automatically gets fetched and loaded. It will use `registerTemplate` under the hood. That's probably the best possible technique for an app the size of brief.

tobowers:
this is great.  And can I still point it at a directory? or is that a tool-chain type thing now?

yaymukund:
Tool-chain thing. The Rails gem will have a way to configure this.. I'm thinking path or extension based, but we haven't decided yet.

tobowers:
so when we merge this it'll not work with brief until that gem is built?

yaymukund:
The `require` stuff (which is what brief currently uses) is [still supported but deprecated](https://github.com/Amicus/end-dash/pull/32/files#diff-a9702bf9ddb61ca29bf4557bfd1b504eR31) so we can push the end-dash changes and get that tested in production as we work on the Rails gem piece.

lib/page_helper.js
```diff
@@ -0,0 +1,24 @@
...
+//Trims leading and ending whitespace, necessary for JQuery creation of DOM elements
+var normalizeTemplateName = function(template) {
+  return template.textContent.replace(/^\s*/, "").replace(/\s*$/, "");
+};
+
+var getTemplatesOnPage = function() {
+  var scriptTags = window.document.getElementsByTagName('script');
+
+  return _.filter(scriptTags, function(scriptElem) {
+    return scriptElem.type == 'text/enddash'
```
tobowers: this was done for performance reasons over using jquery?

yaymukund: No, this was done because I missed it while doing the final cleanup. You're correct— we should probably just use jQuery since we use it for everything else in end-dash.

test/end-dash.js
```diff
@@ -0,0 +1,79 @@
...
+      var model = new Backbone.Model({gooseName: 'Snowy'}),
+          template = EndDash.getTemplate('goose', model);
+
+      expect(template.el.text()).to.be('Snowy');
+      model.set('gooseName', 'Goosematix');
+      expect(template.el.text()).to.be('Goosematix');
+    });
+  });
+
+  describe('loadTemplatesFromPage', function() {
```
tobowers, line 47: should we test that the _content_ of the script tags gets loaded?

yaymukund, line 47: That's what [this last test (L74)](https://github.com/Amicus/end-dash/pull/32/files#diff-0a4b832f090abca5aed7cd1424a3c755R75) is trying to do, though I guess it's focused on whether the template was bound. We can add one that just checks `template.el`.

tobowers, line 47: now that I see that I'm willing to merge without the test... should I go? oh - jquery thing... forgot about that. Shall I wait or go?

yaymukund:
@tobowers Alright, this should be ready for another review. Fixed the jQuery thing.

tobowers:
i really like these changes... merging.

PR #33, opened 2013-10-21
Bump version in preparation to merge with brief.

yaymukund:
@tobowers cc @devmanhinton 

v0.3.0

Breaking changes:
- `EndDash.getTemplate` now returns an instance of a Template rather
  than the class itself.


tobowers:
I think this needs that in bower.json as well

yaymukund:
@tobowers fixed!

PR #36, opened 2013-10-21
Expose setGetViewFn to the end user

devmanhinton:
@yaymukund This needs to be in here. 

Not sure if require the view a second time is okay? (its fast since node caches the require).

Thoughts?

### DIFF
--- a/lib/end-dash.js
+++ b/lib/end-dash.js
@@ -13,7 +13,8 @@ var _ = require('underscore'),
       require('./reactions/conditional'),
       require('./reactions/view'),
       require('./reactions/debugger')
-    ];
+    ],
+    viewReaction = require('./reactions/view'); //required a second time to use directly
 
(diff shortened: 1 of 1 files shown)


devmanhinton:
@yaymukund Calling it a night at least for now but this is necessary to move forwards with the node example app

yaymukund:
Do explain...

devmanhinton:
Look at the code but 

``` javascript
<div data-view="whatever"></div> 
```

relies on the View reaction which relies on a function that is by default undefined, getView. 

In brief, base-backbone deals with setting this up. It does ViewReaction.setGetView(fn...)

So in order to have EndDash work with views we need to expose this function to them.

tobowers:
Is there a way to have a sensible default that will try a require?
- Topper

devmanhinton:
@tobowers Sure, let me write a draft of that and you can let me know what you think

PR #37, opened 2013-10-22
Add a registerView method that lets you register a view.

yaymukund:
@tobowers @devmanhinton 

``` javascript
// BEFORE
// Register a method that returns the view.
ViewReaction.setGetView(function(name) {
  return require(name+'.js.ed');
});

// AFTER
// Register the view itself.
EndDash.registerView('user', UserView);
```

And then use it:

``` html
<div data-view="user">
  <!-- user... -->
</div>
```

### DIFF
--- a/lib/end-dash.js
+++ b/lib/end-dash.js
@@ -1,6 +1,7 @@
 var _ = require('underscore'),
     Parser = require('./parser'),
     TemplateStore = require('./template_store'),
+    ViewStore = require('./view_store'),
     PageHelper = require('./page_helper'),
     reactions = [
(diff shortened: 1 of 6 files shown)


tobowers:
this looks good to me. @devmanhinton you like it too?

devmanhinton:
@tobowers Yep! Sorry for the latish response. @yaymukund Cool work abstracting the backwards capaiblity into the exports.getView function (in view_store.js)

devmanhinton:
Also, closing my initial hacky way pull request :p

PR #38, opened 2013-10-22
Move all test templates to test/support/templates to clean up.

yaymukund:
@tobowers cc @devmanhinton 
Does what it says on the tin.

:dancer:


devmanhinton:
Cool stuff!

PR #39, opened 2013-10-23
Fix diff backbone issue

devmanhinton:
@yaymukund cc @tobowers 

This is a fix for when the Backbone EndDash uses is different then the Backbone that the client's app uses.

Ie., EndDash's Backbone === Client's Backbone // returns false

I ran into this with my node application which was a hella of a pain to track down but was rewarding once I did!

An important note is, though I do not think this will ever come up, adding a backbone model to a collection from a different backbone will not work properly (this is spec-ed). 

===== Below is fun random fact ==== no need to read ==== repeat fun & AWESOME but not necessary
(More details on last sentence). Backbone collections do ~ if not Backbone.Model || Collection convert it to use Backbone style events (so collections can listen for changes). That means take all top level attributes -> attribute object and do ModelPassedIn.attributes = attributes. With a BackboneModel from a different Backbone this overwrites its attributes with a nested version of itself. In conclusion. Collection.get(model).attributes.attributes is actually the model's attributes. Fun times and likely unncessary.

### DIFF
--- a/lib/reactions/model.js
+++ b/lib/reactions/model.js
@@ -9,7 +9,7 @@ var ModelReaction = Reaction.extend({
   init: function(next, reload) {
     var newModel = this.get(this.modelName);
     if(!(newModel instanceof Backbone.Model || newModel instanceof Backbone.Collection ||
-          this.isEndDashCompatbile(newModel))) {
+          this.isEndDashCompatible(newModel))) {
       newModel = new Backbone.Model(newModel || {});
     }
 
@@ -18,7 +18,7 @@ var ModelReaction = Reaction.extend({
     next(reload);
   },
 
-  isEndDashCompatbile: function(model) {
+  isEndDashCompatible: function(model) {
     return (typeof model.on == 'function' &&
             typeof model.once == 'function' &&
             typeof model.set == 'function' &&
--- a/lib/reactions/model.js
+++ b/lib/reactions/model.js
@@ -2,23 +2,23 @@ var Reaction = require("../reaction")
   , rules = require("../rules")
   , path = require("path")
   , Backbone = require("backbone")
-  , _ = require("underscore")
+  , _ = require("underscore");
 
 var ModelReaction = Reaction.extend({
 
   init: function(next, reload) {
-    var newModel = this.get(this.modelName)
+    var newModel = this.get(this.modelName);
     if(!(newModel instanceof Backbone.Model || newModel instanceof Backbone.Collection ||
-          this.isEndDashCompatabile(newModel))) {
-      newModel = new Backbone.Model(newModel || {})
+          this.isEndDashCompatbile(newModel))) {
+      newModel = new Backbone.Model(newModel || {});
     }
 
-    this.el.data("model", newModel)
-    this.stack.push(newModel)
-    next(reload)
+    this.el.data("model", newModel);
+    this.stack.push(newModel);
+    next(reload);
   },
 
-  isEndDashCompatabile: function(model) {
+  isEndDashCompatbile: function(model) {
     return (typeof model.on == 'function' &&
             typeof model.once == 'function' &&
             typeof model.set == 'function' &&
@@ -28,16 +28,16 @@ var ModelReaction = Reaction.extend({
 
   observe: function(next) {
     this.change(this.modelName, function(model) {
-      this.init(next, true)
-    }, this)
+      this.init(next, true);
+    }, this);
     if(this.el.is("form")) {
       this.uiEvent("submit", function(e) {
-        var objectToSave = this.stack[this.stack.length - 1]
+        var objectToSave = this.stack[this.stack.length - 1];
         if (objectToSave.save && objectToSave.save.apply) {
-          objectToSave.save()
+          objectToSave.save();
         }
-        e.preventDefault()
-      }, this)
+        e.preventDefault();
+      }, this);
     }
   }
 
@@ -56,4 +56,4 @@ var ModelReaction = Reaction.extend({
   }
 })
 
-module.exports = ModelReaction
+module.exports = ModelReaction;
--- a/lib/reactions/model.js
+++ b/lib/reactions/model.js
@@ -26,7 +26,7 @@ var ModelReaction = Reaction.extend({
             typeof model.once == 'function' &&
             typeof model.set == 'function' &&
             typeof model.get == 'function'
-           )
+           );
   },
 
   observe: function(next) {
@@ -48,15 +48,15 @@ var ModelReaction = Reaction.extend({
   selector: "[class]",
 
   reactIf: function(el) {
-    return rules.model(el)
+    return rules.model(el);
   },
 
   parse: function(el, state) {
     return {
       modelName: rules.model(el),
       currentDirectory: state.currentDir()
-    }
+    };
   }
-})
+});
 
 module.exports = ModelReaction;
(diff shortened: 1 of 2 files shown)


test/non_backbone_integration.js
```diff
@@ -0,0 +1,184 @@
...
+    model.set('persisted', 'Changed')
+    expect($('.persisted-').html()).to.be('Changed')
+  })
+})
+
+describe("With a backbone model", function(){
+  beforeEach(function(){
+    var backModel = new Backbone.Model()
+  })
+  it("Should work as normal", function(){
```
yaymukund: This test description should be more specific

lib/reactions/model.js
```diff
@@ -18,6 +18,14 @@ var ModelReaction = Reaction.extend({
     next(reload)
   },
 
+  isEndDashCompatabile: function(model) {
```
yaymukund: should be `Compatible`

yaymukund:
As I said before, we should be using semicolons consistently in new code.

test/non_backbone_integration.js
```diff
@@ -0,0 +1,184 @@
...
+    expect($('.persisted-').html()).to.be('Chelsa Piers')
+    model.set('persisted', 'Changed')
+    expect($('.persisted-').html()).to.be('Changed')
+  })
+})
+
+describe("With a backbone collection from a different backbone object", function(){
+  beforeEach(function(){
+    this.collection = new BackboneClone.Collection()
+    this.templateString = "<div class='users-'><div class ='user-'><div class=' #{hook} persisted-'></div></div></div>"
```
yaymukund: Split up this string over multiple lines so it's easier to read

lib/reactions/model.js
```diff
@@ -18,6 +18,14 @@ var ModelReaction = Reaction.extend({
...
+    return (typeof model.on == 'function' &&
```
yaymukund, line 25: Still not comfortable with this implementation but I think it's good enough for now. I'll open up an issue later with some thoughts on how to do this correctly.

test/non_backbone_integration.js
```diff
@@ -0,0 +1,184 @@
...
+
+delete require.cache[require.resolve('Backbone')]
+var BackboneClone = require('Backbone') // Internally we use !(model instanceof Backbone.Model || ... instanceof Backbone.Collection)
+                                        // If the Backbone for EndDash is different then the clients, we want to handle this case.
+                                        // See lib/reactions/model.js for fix.
+
+describe("With a backbone model from a different backbone object", function(){
+  it("binding it to a template should work normally", function(){
+    var model = new BackboneClone.Model({persisted: "Chelsa Piers"})
+      , template = generateTemplate({user: model}, "<div class='user-'><div class='persisted-'></div></div>")
```
yaymukund: Please split this string over multiple lines so it's easier to read. Same goes for all the `generateTemplate` calls in this PR.

test/non_backbone_integration.js
```diff
@@ -0,0 +1,189 @@
...
+  , generateTemplate = require("./support/generate_template")
+  , Backbone = require('Backbone');
+
+delete require.cache[require.resolve('Backbone')];
+var BackboneClone = require('Backbone');// Internally we use !(model instanceof Backbone.Model || ... instanceof Backbone.Collection)
+                                        // If the Backbone for EndDash is different then the clients, we want to handle this case.
+                                        // See lib/reactions/model.js for fix.
+
+describe("With a backbone model from a different backbone object", function(){
+  it("binding it to a template should work normally", function(){
```
yaymukund: These tests are okay, but you can make them more semantic. For example, here's how I'd rewrite this test:

``` javascript
var user = new Backbone.Model({name: 'Chelsa Piers'}),
    template = generateTemplate({user: user}, markup);

expect($('.name').text()).to.be('Chelsa Piers');
user.set('name', 'Larry Gonick');
expect($('.name').text()).to.be('Larry Gonick');
```

Realize that people sometimes skim tests in order to familiarize themselves with the library, so it's doubly important to make them easy to read.

test/non_backbone_integration.js
```diff
@@ -0,0 +1,189 @@
...
+      , template = generateTemplate({user: model}, "<div class='user-'><div class='persisted-'></div></div>");
+
+    expect($('.persisted-').html()).to.be('Chelsa Piers');
+    model.set('persisted', 'Changed');
+    expect($('.persisted-').html()).to.be('Changed');
+  })
+})
+
+describe("With a backbone model", function(){
+  beforeEach(function(){
```
yaymukund: `backModel` never seems to be used.

test/non_backbone_integration.js
```diff
@@ -0,0 +1,189 @@
...
+    var model = new Backbone.Model({persisted: "Chelsa Piers"})
+      , template = generateTemplate({user: model}, "<div class='user-'><div class='persisted-'></div></div>");
+
+    expect($('.persisted-').html()).to.be('Chelsa Piers');
+    model.set('persisted', 'Changed');
+    expect($('.persisted-').html()).to.be('Changed');
+  })
+})
+
+describe("With a backbone collection from a different backbone object", function(){
```
yaymukund: So much copy-pasted code here. Can you please dry it up using `beforeEach` or something?

lib/reactions/model.js
```diff
@@ -2,34 +2,42 @@ var Reaction = require("../reaction")
...
+    this.stack.push(newModel);
+    next(reload);
+  },
+
+  isEndDashCompatbile: function(model) {
+    return (typeof model.on == 'function' &&
+            typeof model.once == 'function' &&
+            typeof model.set == 'function' &&
+            typeof model.get == 'function'
+           )
```
yaymukund: `;`

:page_facing_up:

lib/reactions/model.js
```diff
@@ -2,34 +2,42 @@ var Reaction = require("../reaction")
...
 
-    this.el.data("model", newModel)
-    this.stack.push(newModel)
-    next(reload)
+    this.el.data("model", newModel);
+    this.stack.push(newModel);
+    next(reload);
+  },
+
+  isEndDashCompatbile: function(model) {
```
yaymukund: Another typo?

yaymukund:
@devmanhinton Let me know when this stuff is fixed and you want me to review again.

devmanhinton:
@yaymukund 

In terms of the spec:

In addition, I moved all the models & templates into a beforeEach top level describe block as well as use JQuery selectors instead of the hook I was using before. Finally, I added a lotta semi-colons. I think this really helps the readability.

In terms of a new bug I found and fixed:

Bug:
Template generated that scopes down to a model or property that is undefined*
lib/reactions/model.js-- calls isCompatible(model)
inside of isCompatible-- calls model.on
--> Exception, cannot call property 'on' of undefined.

Fix:
Add if(!model) return false; to lib/reactions/model.js:isCompatible (commented in this PR)

*note: I'm not sure of the exact cause but is from brief script edit view (creating a new script)

lib/reactions/model.js
```diff
@@ -2,34 +2,45 @@ var Reaction = require("../reaction")
...
-    next(reload)
+    this.el.data("model", newModel);
+    this.stack.push(newModel);
+    next(reload);
+  },
+
+  isEndDashCompatible: function(model) {
+    if (!model) {
+      return false;
+    }
```
devmanhinton, line 24: Above is check if model passed in is undefined

devmanhinton:
Missed committing a few semi colons in lib/reaction/model.js. Let me do that after our meeting before merging.

bglusman:
@yaymukund this seems good to merge now, but I'll let you pull trigger since you were reviewer.

yaymukund:
The tests could use some cleanup, but this has been sitting around long enough so I'm going to merge.

PR #41, opened 2013-10-25
rootElements method and tests

bglusman:
@yaymukund this doesn't 'handle' the multiple elements yet so I guess branch is mis-named, but it's the method to at least raise an error when it's count is > 1 for now, and should make handling them easier by treating each of it's return values as a separate template.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -28,6 +28,15 @@ var Parser = module.exports = function(markup, opts) {
 
 _.extend(Parser.prototype, Backbone.Events)
 
+var _htmlFromNode = function(node) {
+  return $($('<body/>').append(node)).html()
+}
+
+Parser.rootElements = function(markup) {
+  nodes = $.parseHTML( markup );
+  return _.map(nodes, function(node) { return _htmlFromNode(node) });
+}
+
 Parser.registerReaction = function(reaction) {
   reactions.push(reaction)
 }
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -23,6 +23,11 @@ var Parser = module.exports = function(markup, opts) {
   }
 
   this.preparse(this.markup)
+
+  if (rootElements(markup).length > 1) {
+    throw new Error('multiple root elements in a template not currently supported - template: '+opts.templateName)
+  }
+
   this.parse(this.markup)
 }
 
@@ -34,7 +39,8 @@ var _htmlFromNode = function(node) {
 
 Parser.rootElements = function(markup) {
   nodes = $.parseHTML( markup );
-  return _.map(nodes, function(node) { return _htmlFromNode(node) });
+  clean_nodes = _.reject(nodes, function(node) { return !$.trim(_htmlFromNode(node)) })
+  return _.map(clean_nodes, function(node) { return _htmlFromNode(node) });
 }
 
 Parser.registerReaction = function(reaction) {
--- a/lib/template_store.js
+++ b/lib/template_store.js
@@ -26,10 +26,6 @@ var _parseTemplate = function(templatePath) {
   })).generate();
 };
 
-var _htmlFromNode = function(node) {
-  return $($('<body/>').append(node)).html()
-}
-
 TemplateStore.isLoaded = function(templatePath) {
   return !!RAW_TEMPLATES[templatePath];
 };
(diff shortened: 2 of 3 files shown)


yaymukund:
Why not do this [in `lib/parser.js`](https://github.com/Amicus/end-dash/blob/6821e98a9e934e1f269ea8533a16b51049500308/lib/parser.js#L11)? That way, we wouldn't have to parse the html in two places.

bglusman:
Yeah, that might make more sense... will play with trying to move it there.
 Also have unpushed changes to raise error on more than 1 root element, but
that actually fails a bunch of other tests...  one, in
backbone_integration, clealry currently has multiple root elements, but
some of the others it's less clear why... tempted to push up as is and
investigate failures later, but dunno.

215-460-9585 - _cell_
_Twitter_: @amicushq

yaymukund:
Cool, let me know when you want another review (GH doesn't send an email if you push changes, so you have to explicitly tag me in a comment).

bglusman:
@yaymukund this does the minimal thing we discussed for now, raising on multiple roots... tempted to keep going and actually handle them but should probably save that for after we've met other goals/make that a stretch goal.  @devmanhinton welcome to take a look too.

yaymukund:
Maybe I'm missing something obvious but why couldn't you do something like?:

``` javascript
// lib/parser.js#11
this.markup = $(markup);
if (this.markup.length !== 1) {
  throw new Error(...);
}
```

bglusman:
Yeah, that actually does work too, I thought of trying that at one point
but forgot...  possible that'd even simplify what I'm doing to return the
list of root strings too, but wanting that to eventually support multiple
root notes may have lead me astray (though as I said, this may have worked
there too, forgot to try after I thought of it myself)

Checked and it doesn't work for the actual parsing, and seems weird to have two different code paths...  to eventually support multiple roots I think this will help, so inclined to leave it as is, but if you strongly prefer we could change that one line to use this.markup.length instead of rootElements(markup).length

devmanhinton:
@bglusman Do not quite follow the code, why trim the nodes white space after creation?

Can we do what @yaymukund suggested:

I'm guessing either way works but the later's functionality and methodology is clear from a first glance.

yaymukund:
> Checked and it doesn't work for the actual parsing

How's that?

devmanhinton:
Whoops read to fast, I'll second Mukund's question

bglusman:
Had to trim for new lines (i think) being detected as extra (but empty)
nodes.  Tried using using _.isEmpty which seems to work in browser on
newlines, but had some issues so went to trim.

Sure, we can do that for the exception raising test...  I misunderstood mukund was quoting me in his question, I meant that I tried using $(markup) instead of $.parseHTML(markup) in my method, and it won't pass my tests that way.  So we can use it, but if we're still hoping to support multiple roots in a template and use the code I did here, it feels weird to detect the separate nodes one way in one place, and another way in another place.

bglusman:
sorry, updated above comment @yaymukund and @devmanhinton but was responding via email so if you're reading there may be confused :-)

yaymukund:
@bglusman err, sorry, I wasn't specific.

Why don't you detect multiple root elements _when you initialize the Parser_? So instead of having a separate function, just do something like:

``` javascript
var Parser = module.exports = function(markup, opts) {
  opts = opts || {};

  this.markup = $(markup);
  if (this.markup.length !== 1) {
    throw new Error(...);
  }
```

I've tested this approach a bit in the browser and it seems to work as expected:

``` javascript
$('<div></div>').length; // 1
$('<div></div><div></div>').length; // 2
```

> we're still hoping to support multiple roots in a template

Hmm, if we want to support multiple roots, it will take more work later. Let's go with detecting multiple roots and throwing an error as simply as possible for now.

devmanhinton:
+1

bglusman:
Yes, @yaymukund as I said, that totally works for now, and if only one of us had thought of it at the beginning I would have done that way and stopped :-)  I could have done faster/sooner doing the same thing with parseHTML(markup).length and probably should have, but I still wasn't sure if we were decided on just raising an error or trying to support it.  Anyway, enough discussion, I'll push up a version doing it your way and we can tackle other questions later, just unsure whether to leave in any of the rest of the code that will make supporting multiple roots later easier if it's not going to be used now. I guess should delete for now and put that in a separate branch can return to if/when we have time for multiple roots.  This might have been way easier to discuss in person than via PR, apologies for confusion.  cc/ @devmanhinton

bglusman:
Closing in favor of https://github.com/Amicus/end-dash/pull/43 for now

PR #43, opened 2013-10-28
simple error on multiple roots

bglusman:
@yaymukund @devmanhinton clean simple leave old branch and close that PR.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -9,6 +9,11 @@ var Parser = module.exports = function(markup, opts) {
   opts = opts || {}
 
   this.markup = $(markup)
+
+  if (this.markup.length > 1) {
+    throw new Error('multiple root elements in a template not currently supported - template: '+opts.templateName)
+  }
+
   this.absolutePath = opts.templateName
   this.reactions = reactions
   this.structure = { children: [] }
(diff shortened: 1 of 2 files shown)


lib/parser.js
```diff
@@ -9,6 +9,11 @@ var Parser = module.exports = function(markup, opts) {
   opts = opts || {}
 
   this.markup = $(markup)
+
+  if (this.markup.length > 1) {
```
yaymukund: How about `!== 1`? Is `0` valid?

bglusman: I don't think we should raise an error for 0, at least not this error, since it would be... misleading, as 0 is not multiple roots.  It seems more likely to cause problems/confusion.  If 0 is invalid AND causes a problem, I think it should be addressed separately.  If it's invalid but works fine, or invalid but never occurs in practice, I think >1 is more intention revealing anyway.

yaymukund: Can you make it say, `"Template 'foo' invalid: a template must have exactly one root node."`? Would that resolve all your questions?

bglusman: Could I guess...  unclear to my why you prefer this so strongly, but I guess it's fine with me, just feels odd.  Rather get it merged in than argue what seems like a tiny difference though, so I guess I'll push up that change if you don't want to merge as is.

devmanhinton:
Cool @bglusman. If you push it up, I'll merge it :)

bglusman:
@devmanhinton should be good now

PR #44, opened 2013-10-28
Collection docs

devmanhinton:
@yaymukund @bglusman I believe this is ready to merge given both of your approval.

It still needs style tweaks but Mukund, you said designers will look this over and do that.

### DIFF
--- a/docs/collection.md
+++ b/docs/collection.md
@@ -1,11 +1,12 @@
 Collections
 ===========
 
-How do I use them?
-===================
+What if I have a colletion of models that each need
+to use the same template?
 
-EndDash will allow you to render a single template
-once for each item in your collection!
+Great! EndDash lets you render a single template
+once for every model with a collection with no
+extra work!
 
 In your HTML
 ============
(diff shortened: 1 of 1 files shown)


yaymukund:
While I'm glad you worked on this and I appreciate that I didn't have examples for you to mimic, I disagree with the decision to merge this. Some problems:
- Typos. (It's not the designers' responsibility to fix your typos)
- Too much documentation! We can't expect users to read this much for a single feature. We should have less text and more examples.
- This is the wrong place to document the "things will update in the template if you change the underlying objects" feature. We don't want to document that for every single end-dash feature...
- I read this once and I'm still not sure what polymorphic keys are meant for. There should be a clearer example.

I'm not spending more time on this right now since it's already been merged, but we should revisit this before we make it public.

sethbannon:
@yaymukund is there a place you're keeping track of the "things we should revisit before we make this public"?

yaymukund:
@sethbannon Not yet. If I don't get done tonight with docs, I'll throw it in a pivotal task or something.

PR #45, opened 2013-10-28
delete require extensions, not used in tests or browser anymore

bglusman:
@yaymukund

### DIFF
--- a/lib/end-dash.js
+++ b/lib/end-dash.js
@@ -41,7 +41,3 @@ EndDash.compile = function(module, filename) {
 
   module.exports = TemplateStore.loadAndParse(templateName, markup);
 };
-
-if (require.extensions && !require.extensions['.ed']) {
-  require.extensions['.ed'] = require.extensions['.js.ed'] = exports.compile;
-}


yaymukund:
Looks good to me, but I totally thought this was getting used... how have you tested this?

If you haven't already, you can test enddash in brief by symlinking `app/assets/javascripts/components/end-dash`

bglusman:
@yaymukund yep, tested by running all brief mocha tests, and poking around with some browser testing...  can't think of anything else to test, but lmk if you think of anything, or @devmanhinton if you have ideas/saw possible implications of removing this.

devmanhinton:
@bglusman Yeah, if you symLinked and it worked fine. Then this is good with me.

I guess that is deprecated, even in brief, given this in base-backbone:index.js

  var loaders = {
    template: function(templateDir) {
      _(listFiles(templateDir)).each(function(path) {
        var noExt = path.replace(/.js.ed(.erb)?$/, "")
        EndDash.registerTemplate(noExt, fs.readFileSync(path, 'utf8'))
      })
      _(listFiles(templateDir)).each(function(path) {
        var noExt = path.replace(/.js.ed(.erb)?$/, "")
        if(window && window.require) {
          window.require.register(noExt + ".js", function(module) {
            templateCompiler(module, path)
          })
        }
      })
    },

templateDir in this case is /app/assets/javascript/new/templates & that is the only place where there are '.js.ed' files. So this is a safe merge.

bglusman:
Yeah, change was fine in brief.  WIll let one of you two click merge button on this since you both seem OK with it.

PR #46, opened 2013-10-28
Add documentation

yaymukund:
@devmanhinton @bglusman 

Added basic documentation for enddash models and changing scope. Every feature should include a simple example. We can also eventually add some more complex examples that combine different features.


docs/enddash-sytax.md
```diff
@@ -0,0 +1,47 @@
+# EndDash Syntax Documentation
+
+## Variables
+
+```html
+<div class="user-">
+  My name is #{firstName} #{lastName}.
+</div>
+```
+
```
devmanhinton: This may be too terse. There is no text besides titles in this documentation.

For ex: Perhaps 

## Variables

' #{variableName}  ' is used to interpolate variables into html tag values

devmanhinton: Also, this is not true. Variable interpolation only works inside of tag attribute values:

``` html
<div class='#{firstName}'>
```

Below is not valid.

``` javascript
<div class='user-'>
    My name is #{firstName}
</div>
```

yaymukund: Thanks for taking a look so quick. I've fixed that example.

Why is this too terse? Do you there's something that people won't understand?

I do think we need someplace to discuss design decisions and make an argument for end-dash over other templating libraries, but that should be in the readme and not the syntax docs.

docs/enddash-sytax.md
```diff
@@ -0,0 +1,47 @@
+# EndDash Syntax Documentation
```
bglusman: typo in filename, syntax not sytax :-)

bglusman:
Looks like a good start, could merge now but will let you fix file name first, maybe i'll see if I can find any other feedback reading closer...

docs/enddash-sytax.md
yaymukund: HAH. Good catch. Thanks. Fixed @bglusman.

docs/enddash-sytax.md
devmanhinton: Excellent.

And yes, though it is hard to tell given how little of this document is written. Many syntax rules are not in here yet. 

We need to at some point clarify on which DOM elements these syntactical rules can be applied to. I would suggest a white list strategy.

Keep writing this doc, push it up, and then I will read over it. It is hard to review without seeing the future picture of what you have planned.

PR #47, opened 2013-10-28
Refactor template store

bglusman:
@yaymukund and/or @devmanhinton this probably isn't ready for merge yet, or if it is probably want to recreate test/end-dash.js in new test style mukund used in his branch, which is in this branch also.  Was unsure how much duplication there was there at a glance and it was pretty clear it was failing only because it's new and not using his new conventions... Maybe I can adapt later if I have time or maybe one of you wants to take a shot since you two wrote it.

Otherwise I think this is a pretty successful port of that branch, but take a look and let me know.


lib/template_store.js
```diff
@@ -3,65 +3,68 @@
...
   }
 };
 
 TemplateStore.loadAndParse = function(templatePath, markup) {
   this.load(templatePath, markup);
-  return this.getTemplateClass(templatePath);
+  return this.getTemplate(templatePath);
 };
 
-TemplateStore.getTemplateClass = function(templatePath) {
```
yaymukund, line 52: Why rename this? It's returning a class, not an instance.

bglusman, line 52: Could be a mistake?  I was mostly applying the same changes from your branch manually, that was one of them.  We can change back if I got wrong or you changed your mind.

yaymukund:
What do you need to do before it's ready for merge?

bglusman:
Mostly I want to adapt/add back in the test/end-dash.js that I deleted because it was failing, can probably manage that myself but thought it might be good during pairing since you reworked test conventions a bit and wanted to make sure I understood all of how and why you changed. I think it could be merged in now though and that can be a seperate PR if you prefer.

bglusman:
@yaymukund sorry didn't tag you in above comment

lib/template_store.js
yaymukund, line 52: This is your code now. Please make sure you understand every line of a diff you're asking me to review (or add comments where you're unsure).

bglusman, line 52: I think we'll need to discuss this statement face to face, not prepared to respond casually in writing because this feels like another communication challenge/failure.

yaymukund:
As i said in the inline comment, please review this diff first. You may want to work from [the diff from the PR (`merge-base..HEAD`)](https://github.com/Amicus/end-dash/pull/28/files) rather than the diff from master (`master..HEAD`).

I haven't had a chance to dive in too deep, but a couple things that don't make sense:
- `test/util.js` contains the same code as `test/support/generate_template.js`
- `test/support/` has a bunch of template files that should be in `test/support/templates`.

bglusman:
Thanks, that's more constructive for me than the inline comment, which...  bothered me a bit :-)

bglusman:
new branch coming

PR #49, opened 2013-10-29
Merge zach inflectionless

devmanhinton:
@yaymukund These are Zach's changes that I merged locally with master.
cc @bglusman 
Collection iteration is no longer achieved via:

``` html
<div class='users-'>
  <div class='user-'>
     //Rest of HTML
  </div>
</div>
```

But rather adding tag attribute 'data-col' or 'data-collection'

So

``` html
<div class='users-' data-col>
  <div>
    //Rest of HTML
  </div>
</div>
```

Main note: this will require changing brief: adding data-col to every collection iteration we do in brief. If you like these changes I will make a new branch in brief and test it by symbolic linking it to my local copy of EndDash.


bglusman:
This will break end-dash all over the place in brief if i understand correctly, but will also make end-dash no longer support collection attributes I think, which invites templates needs and uses as we determined yesterday...  unless it can resolve whether the attribute is on the model or the collection somehow?

devmanhinton:
How so? 

I do not think that is the case.

Can you give me one or two examples of where this will break brief?

The change I see is

From:

``` html
<div class='users-'>
  <div class='user-'>
   //Other HTML
  </div>
</div>
```

To:

``` html
<div class='users-' data-col>
   <div>
   //Other HTML
   </div>
</div>
```

1) Add data-col
2) Remove 'user-'
3) Update CSS rules

How this would change collection attributes? ( I am not familiar with them)

bglusman:
@devmanhinton look at js/new/templates/invites/new.js.ed in brief...  invitesSent- and invitesComplete- are properites on the collection... if you get rid of (what would be the invite-) section and make it implicit, how will invitesSent- and invitesComplete- resolve to the collection attributes, vs instance attributes?

The other stuff in brief is small, but very numerous, and potentially hard to be sure we've gotten them all... I'd be cool if we supported both methods and logged a deprecation warning about the old method when used, if we can also resolve the collection attributes issue.

bglusman:
also I think data-collection would be better than data-col as col might also be short for column :-)

devmanhinton:
So you can use data-collection or data-col :)

And that would work fine. We could leave all that as is.

There is no ' invite- ' in that template. There is ' newInvite-', that is not collection iteration though. That is the property of the object being passed in.
https://github.com/Amicus/brief/blob/master/app/assets/javascripts/new/routes/invites.js.module.coffee#L25

There is actually no collection iteration in that template :)

bglusman:
new branch coming

PR #51, opened 2013-10-29
Refactor view store non singleton

bglusman:
@yaymukund may conflict with your template store refactor but should be pretty good otherwise.

### DIFF
--- a/lib/reactions/view.js
+++ b/lib/reactions/view.js
@@ -3,12 +3,12 @@ var Reaction = require("../reaction"),
     rules = require("../rules"),
     Backbone = require("backbone"),
     _ = require("underscore"),
-    ViewStore = require('../view_store');
+    viewStore = require('../end-dash').viewStore;
 
 var ViewReaction = Reaction.extend({
   init: function(next) {
     var oldView = this.el.data("view"),
-        viewClass = ViewStore.getView(this.viewName),
+        viewClass = viewStore.getView(this.viewName),
         view, opts;
 
     if (typeof oldView === "object") {
@@ -45,7 +45,7 @@ var ViewReaction = Reaction.extend({
     return { viewName: rules.view(el) };
   },
 
-  setGetView: ViewStore.setCustomGetView
+  setGetView: viewStore.setCustomGetView
 });
 
 module.exports = ViewReaction;
--- a/lib/reactions/view.js
+++ b/lib/reactions/view.js
@@ -3,12 +3,12 @@ var Reaction = require("../reaction"),
     rules = require("../rules"),
     Backbone = require("backbone"),
     _ = require("underscore"),
-    viewStore = require('../end-dash').viewStore;
+    EndDash = require('../end-dash');
 
 var ViewReaction = Reaction.extend({
   init: function(next) {
     var oldView = this.el.data("view"),
-        viewClass = viewStore.getView(this.viewName),
+        viewClass = EndDash.viewStore.getView(this.viewName),
         view, opts;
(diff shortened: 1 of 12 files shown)


lib/view_store.js
```diff
@@ -1,23 +1,30 @@
...
-exports.getView = function(name) {
-  if (VIEWS[name]) {
-    return VIEWS[name];
+  this.load = function(name, View) {
+    VIEWS[name] = View;
+  };
 
-  } else if (typeof getView === 'function') {
-    return getView(name);
+  this.getView = function(name) {
```
yaymukund: You're declaring functions on the instance directly. This means that we'll create a copy of `load`, `getView`, `setCustomGetView` every time we call `new ViewStore()`. Instead, you can declare functions on the prototype:

``` js
function ViewStore() {
  this.VIEWS = {};
  this.getView = null;
}

ViewStore.prototype.load = function(name, View) {
  this.VIEWS[name] = View;
};

ViewStore.prototype.getView = function(name) { ... }
```

bglusman: Ahh, right, forgot, thanks...

bglusman: Curious about the this.getView = null though... kind of follow what with setCustomGetView, but little fuzzy why need explicit null, won't it be null on instance by default?

Anyway, so far get 3 errors in tests when I set that to null, and one without it, as if VIEWS were undefined, but thx for review/help.

bglusman:
@yaymukund merged in your templateStore refactor because they were interdependent/doing much of the same stuff...  @devmanhinton thanks for your help, tests now pass, hopefully this is ready for CR and merge now.

lib/reactions/view.js
```diff
@@ -3,12 +3,12 @@ var Reaction = require("../reaction"),
...
     _ = require("underscore"),
-    ViewStore = require('../view_store');
+    EndDash = require('../end-dash');
 
 var ViewReaction = Reaction.extend({
   init: function(next) {
     var oldView = this.el.data("view"),
-        viewClass = ViewStore.getView(this.viewName),
+        viewClass = EndDash.viewStore.getView(this.viewName),
         view, opts;
```
devmanhinton, line 12: We need to decide wether to do:

``` javascript
EndDash.viewStore.getView(...
```

Or define a wrapper on EndDash and do:

``` javascript
EndDash.getView(this.viewName..
```

Just need to pick one and stay consistent. This is an open question for you, Mukund, and me.

yaymukund, line 12: We should do `EndDash.viewStore.getView(...)` within EndDash. There might eventually be `viewStore` methods that we'll want to use internally, but we don't want to expose on `EndDash`.

devmanhinton:
@bglusman  This seems good to me. Mukund's branch is merged into this one. That branch is about to be updated with some changes. 

Maybe:

1) Mukund fixes his branch, pushes -> we CR & merge
2) Make things mergeable
      --- a) View_store
      --- b) New-collection-iteration
3) Meanwhile Mukund can CR
4) Back & forth code review/edits on 2a & 2b
5) Merge 2a & 2b

Then we have:
1) New collection iteration
2) More prototypal and non-singleton modules

Finally:

1) PR new-collection-iteration-deprecation (Mukund CR, we refactor based on comments)
2) Refactor the parser/template_store/end-dash

Hopefully all by EOD

lib/reactions/view.js
devmanhinton, line 12: Sure, leaving it this way for now

PR #52, opened 2013-10-30
The TemplateStore uses prototypal inheritance, is no longer singleton.

yaymukund:
@bglusman, @devmanhinton 

This branch makes it so we use prototypal inheritance and create `templateStore` instances (v. having a singleton `TemplateStore`).

``` js
// Before:
exports.create = function() {
  return {
    getTemplate: function() {...},
    isLoaded: function() {...},
    TEMPLATES: {},
    // etc.
  }
};

// After:
function TemplateStore() {
  this.TEMPLATES = {};
  this.RAW_TEMPLATES = {};
}

TemplateStore.prototype.isLoaded = function() {
  // can use this.TEMPLATES and this.RAW_TEMPLATES
}
```

### DIFF
--- a/lib/end-dash.js
+++ b/lib/end-dash.js
@@ -1,8 +1,9 @@
 var _ = require('underscore'),
     Parser = require('./parser'),
     TemplateStore = require('./template_store'),
-    ViewStore = require('./view_store'),
+    templateStore = new TemplateStore(),
     PageHelper = require('./page_helper'),
+    ViewStore = require('./view_store'),
     reactions = [
       // The load order matters here...
       require('./reactions/partial'),
@@ -18,26 +19,29 @@ var _ = require('underscore'),
 
 _.each(reactions, Parser.registerReaction);
 
-module.exports = EndDash = {};
+exports.templateStore = templateStore;
 
-EndDash.bootstrap = PageHelper.loadFromPage;
-EndDash.registerTemplate = TemplateStore.load;
-EndDash.getTemplateClass = TemplateStore.getTemplateClass;
-EndDash.registerView = ViewStore.load;
+exports.bootstrap = PageHelper.loadFromPage;
 
-EndDash.getTemplate = function(templatePath, model) {
+exports.registerTemplate = _.bind(templateStore.load, templateStore);
+
+exports.getTemplateClass = _.bind(templateStore.getTemplateClass, templateStore);
+
+exports.registerView = ViewStore.load;
+
+exports.getTemplate = function(templatePath, model) {
   var Template = this.getTemplateClass(templatePath);
   return new Template(model);
-}
+};
 
-// This is a hack to make requiring a file that ends in .js.ed to return a
-// EndDash compiled version of that file. We need to expose it publicly,
-// because we have code that relies on it, but it's deprecated and not
-// officially a part of the public API.
-EndDash.compile = function(module, filename) {
+exports.compile = function(module, filename) {
+  // This is a hack to make requiring a file that ends in .js.ed to return a
+  // EndDash compiled version of that file. We need to expose it publicly,
+  // because we have code that relies on it, but it's deprecated and not
+  // officially a part of the public API.
   var fs = require('fs'),
       markup = fs.readFileSync(filename, 'utf8'),
       templateName = filename.replace(/\.js\.ed(\.erb)?$/, '');
 
-  module.exports = TemplateStore.loadAndParse(templateName, markup);
+  module.exports = templateStore.loadAndParse(templateName, markup);
 };
--- a/lib/page_helper.js
+++ b/lib/page_helper.js
@@ -1,5 +1,5 @@
-var _ = require('underscore')
-  , templateStore = require('./template_store');
+var _ = require('underscore'),
+    EndDash = require('./end-dash');
 
 // Only load HTML do not parse until template is requested
 // If this is changed be aware, order matters. If templates with
@@ -15,6 +15,6 @@ exports.loadFromPage = function () {
       throw new Error("Script tags of type text/enddash must have a 'name' attribute");
     }
 
-    templateStore.load($el.attr('name'), markup);
+    EndDash.templateStore.load($el.attr('name'), markup);
   });
 };
--- a/lib/reactions/collection.js
+++ b/lib/reactions/collection.js
@@ -1,10 +1,10 @@
 var Parser = require("../parser")
-  , TemplateStore = require('../template_store')
+  , EndDash = require('../end-dash')
   , Reaction = require("../reaction")
   , inflection = require("inflection")
   , _ = require("underscore")
   , get = require("../util").get
-  , rules = require("../rules")
+  , rules = require("../rules");
 
 var CollectionReaction = Reaction.extend({
 
@@ -117,7 +117,7 @@ var CollectionReaction = Reaction.extend({
       var child = $(element)
         , whenValue = rules.polymorphicValue(child)
         , templatePath = _.last(state.pathStack)
-        , Template = TemplateStore.loadAndParse(templatePath+whenValue, child);
+        , Template = EndDash.templateStore.loadAndParse(templatePath+whenValue, child);
 
       templates[whenValue] = Template;
       child.remove()
--- a/lib/template_store.js
+++ b/lib/template_store.js
@@ -2,66 +2,62 @@
 // and then returning template objects (lazily).
 var _ = require('underscore'),
     path = require('path'),
-    Parser = require('./parser'),
-    TemplateStore = {},
-    RAW_TEMPLATES = {},
-    TEMPLATES = {};
+    Parser = require('./parser');
 
 var pathToName = function(templatePath) {
   return path.normalize(templatePath);
 };
 
-var _parseTemplate = function(templatePath) {
+function TemplateStore() {
+  this.RAW_TEMPLATES = {};
+  this.TEMPLATES = {};
+};
+
+TemplateStore.prototype._parseTemplate = function(templatePath) {
   var name = pathToName(templatePath);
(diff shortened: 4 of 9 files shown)


yaymukund:
bump :ocean:

lib/end-dash.js
```diff
@@ -18,26 +19,29 @@ var _ = require('underscore'),
 
 _.each(reactions, Parser.registerReaction);
 
-module.exports = EndDash = {};
+exports.templateStore = templateStore;
```
bglusman, line 22: Was confused at first that we no longer have an "EndDash" object in the EndDash file... I understand how it works, just putting it's separate functions directly on exports, but is it worth having an EndDash object internally for clarity, or was there a specific reason you got rid of that?

Also, unsure on equivalence etc of module.exports vs just plain exports, I guess both work/are equivalent, but feel like we should be consistent stylistically with one of the other.

lib/end-dash.js
```diff
@@ -1,8 +1,9 @@
 var _ = require('underscore'),
     Parser = require('./parser'),
     TemplateStore = require('./template_store'),
-    ViewStore = require('./view_store'),
+    templateStore = new TemplateStore(),
```
devmanhinton, line 4: Would it make more sense to move this line down after 18?

Every other local variable declaration here is a module inclusion. 

Once we have the view store as prototypal as well then we can break down declarations in the this file to:

a) Load other modules
b) Register reactions
c) Create instances of view & template stores
d) Setup the exports for EndDash

``` javascript
var _ = require('underscore'),
    Parser = require('./parser'),
    TemplateStore = require('./template_store'),
    PageHelper = require('./page_helper'),
    ViewStore = require('./view_store'),
    reactions = [
      // The load order matters here...
      require('./reactions/partial'),
      require('./reactions/scope'),
      require('./reactions/collection'),
      require('./reactions/model'),
      require('./reactions/attribute'),
      require('./reactions/variable'),
      require('./reactions/conditional'),
      require('./reactions/view'),
      require('./reactions/debugger')
    ];

_.each(reactions, Parser.registerReaction);

var viewstore = new ViewStore(),
    templatestore = new TemplateStore();

//Rest of File

```

lib/page_helper.js
```diff
@@ -15,6 +15,6 @@ exports.loadFromPage = function () {
       throw new Error("Script tags of type text/enddash must have a 'name' attribute");
     }
 
-    templateStore.load($el.attr('name'), markup);
+    EndDash.templateStore.load($el.attr('name'), markup);
   });
```
devmanhinton, line 19: This can just be:

``` javascript
EndDash.registerTemplate($el.attr('name'), markup);
```

Since 

``` javascript
exports.registerTemplate = _.bind(templateStore.load, templateStore)
```

lib/reactions/collection.js
```diff
@@ -117,7 +117,7 @@ var CollectionReaction = Reaction.extend({
       var child = $(element)
         , whenValue = rules.polymorphicValue(child)
         , templatePath = _.last(state.pathStack)
-        , Template = TemplateStore.loadAndParse(templatePath+whenValue, child);
+        , Template = EndDash.templateStore.loadAndParse(templatePath+whenValue, child);
 
       templates[whenValue] = Template;
       child.remove()
```
devmanhinton, line 123: What about accessing material not in your module through end-dash.js

Change it to something like:

``` javascript
, templateName = templatePath + whenValue

EndDash.registerTemplate(templateName, child);
templates[whenValue] = EndDash.getTemplateClass(templateName);
```

lib/template_store.js
```diff
@@ -2,66 +2,62 @@
...
 };
 
-var _parseTemplate = function(templatePath) {
+function TemplateStore() {
+  this.RAW_TEMPLATES = {};
+  this.TEMPLATES = {};
+};
+
+TemplateStore.prototype._parseTemplate = function(templatePath) {
   var name = pathToName(templatePath);
```
devmanhinton, line 17: Why the underscore?

test/partials.js
```diff
@@ -24,7 +23,7 @@ describe("A template with partials", function() {
     }
 
     _(templates).each(function(template) {
-      TemplateStore.load(template, fs.readFileSync(__dirname + template).toString())
+      EndDash.templateStore.load(template, fs.readFileSync(__dirname + template).toString())
     })
```
devmanhinton, line 27: Just.

``` javascript
EndDash.registerTemplate(....
```

lib/end-dash.js
```diff
@@ -18,26 +19,29 @@ var _ = require('underscore'),
...
 
-EndDash.bootstrap = PageHelper.loadFromPage;
-EndDash.registerTemplate = TemplateStore.load;
-EndDash.getTemplateClass = TemplateStore.getTemplateClass;
-EndDash.registerView = ViewStore.load;
+exports.registerTemplate = _.bind(templateStore.load, templateStore);
 
-EndDash.getTemplate = function(templatePath, model) {
+exports.getTemplateClass = _.bind(templateStore.getTemplateClass, templateStore);
+
```
devmanhinton, line 29: I thought we were not exposing getTemplateClass to the client?

Is this just for brief?

lib/template_store.js
```diff
@@ -2,66 +2,62 @@
...
 
 var pathToName = function(templatePath) {
   return path.normalize(templatePath);
 };
 
-var _parseTemplate = function(templatePath) {
+function TemplateStore() {
+  this.RAW_TEMPLATES = {};
+  this.TEMPLATES = {};
+};
```
devmanhinton, line 14: Since this is now prototypal, do these need to be all caps?

lib/end-dash.js
yaymukund, line 4: Ah, sure, I'll change this.

lib/page_helper.js
yaymukund, line 19: Good call.

lib/reactions/collection.js
yaymukund, line 123: I don't understand this comment. `loadAndParse(name, child)` basically calls `registerTemplate` and `getTemplateClass` internally, so that code is functionally equivalent.

test/partials.js
yaymukund, line 27: Agreed, good catch.

lib/end-dash.js
yaymukund, line 29: We can discuss removing this in a separate issue. I want this branch to not affect the public interface.

I agree that it would be nice to not expose `getTemplateClass` but I _think_ it's used by something (I don't remember exactly what).

lib/end-dash.js
yaymukund, line 22: This is related to the "empty EndDash object bug" we were experiencing yesterday. Basically, the code goes like:

``` js
// end-dash.js
require('./reactions/collection');

// now in collection.js:
// The "EndDash" variable becomes the "module.exports" object!
EndDash = require('../end-dash');

// collection.js finished processing. Now back to end-dash.js:
module.exports = EndDash = {};
EndDash.templateStore = templateStore;
EndDash.registerTemplate = ...;
```

Do you see the problem? `lib/reactions/collection.js` runs `require('../end-dash')`, retrieving an empty `module.exports` object. Then, `end-dash.js` overwrites that module.exports object and adds methods. But `collection.js` is still referencing the old (empty) `module.exports` object!

This is a sneaky bug and my explanation is probably not enough. If you're still confused, please let's talk in person and I can explain it more clearly!

lib/end-dash.js
devmanhinton, line 29: It is used in brief in ~ 3 places.

Pretty sure we had a deprecated comment on this a while ago. Not a huge issue for now but like you said, something to think about.

lib/reactions/collection.js
devmanhinton, line 123: We have a mix of accessing TemplateStore both through functions bound to templateStore 

``` javascript
exports.registerTemplate = _.bind(templateStore.load, templateStore)
```

And through the templateStore object

``` javascript
EndDash.templateStore.loadAndParse(templatePath+whenValue, child);
```

It would be nice to stick to one method as much as possible for overall clarity of the EndDash repo. 

Does that make sense?

yaymukund, line 123: Ahh, thanks for the clarification. I agree.

lib/end-dash.js
devmanhinton, line 22: Mukund, this implies:

In One file:

``` javascript
//In File 1
window.file1Require = require('exampleFile')
```

In Another:

``` javascript
//In File 2
window.file2Require = require('exampleFile')
```

In any file loaded on the page:

``` javascript
window.file2Require === window.file1Require //returns false
```

No?

yaymukund, line 22: If `exampleFile` changes the `module.exports` object after it gets required, then yeah.

devmanhinton, line 22: Ahhhhhh I see gotcha. Thanks.

devmanhinton:
@yaymukund Push up fixes?

Writing a branch to clean up parser/template_store. These changes would help.

yaymukund:
@devmanhinton, feel free to branch from this branch. I'll help you resolve any git issues when it's ready for PR. Is that good enough?

devmanhinton:
Closing in favor of #51 which has these changes and more within it

yaymukund:
@devmanhinton This should stay open, get fixed, and get merged first. If you want to build on top of it, that's fine, but then your new branch is blocked on this branch getting merged.

devmanhinton:
@yaymukund  No. The new branch #51  already has all changes from this branch merged in + fixes of the errors (style changes). Can you please close again since you re-opened?

devmanhinton:
@yaymukund I will merge this and then other one.

PR #55, opened 2013-10-31
New collection iteration

devmanhinton:
@yaymukund cc @bglusman 

This commit changes collection reactions' syntax & name.

Collection reactions are now called iteration reactions and use
data-each, as opposed to a parent with a plural name of their own class name.

Use:

HTML:

``` html
<div data-each>
  <div>
      Hello my name is <span class='firstName-'></span>
  </div>
</div>
```

JS:

``` javascript
var collection = new Backbone.Collection([new Backbone.Model({firstName: "Brian"}),
                       new Backbone.Model({firstName: "Devon"}),
                       new Backbone.Model({firstName: "Mukund"}) ])

EndDash.getTemplate('hello', collection)
```

Output:

```
Brian
Devon
Mukund
```

We have another branch new-collection-iteration-deprecation which has these changes & backwards compatibly with collection reactions.


yaymukund:
Why did you change `collection` to `iteration`?

lib/reactions/iteration.js
```diff
@@ -0,0 +1,118 @@
+var Parser = require("../parser")
+  , TemplateStore = require('../template_store')
```
yaymukund: Don't start lines with commas.

lib/reactions/iteration.js
```diff
@@ -0,0 +1,118 @@
...
+  , rules = require("../rules");
+
+var IterationReaction = Reaction.extend({
+
+  addAll: function(models, index) {
+    index = index || 0;
+    if(_.isArray(models)) {
+      models = _(models);
+    }; //Hit only if array is passed in and iterated without scoping
+    models.each(function(item, index) {
```
yaymukund: `models.each(this.add, this)`?

lib/reactions/iteration.js
```diff
@@ -0,0 +1,118 @@
+var Parser = require("../parser")
```
yaymukund: I think the name change should be its own commit because it makes this really hard to code review. Is there some way to get a diff of just what has changed in this file?

lib/reactions/iteration.js
```diff
@@ -0,0 +1,118 @@
...
+      , children;
+
+    template.template.data("model", model);
+
+    if(typeof index === "undefined") {
+      this.templates.push(template);
+      this.el.append(template.template);
+    } else if(index === 0) {
+      this.templates.unshift(template);
+      this.el.prepend(template.template);
```
yaymukund: We should be using `template.el` instead of `template.template` in new code.

lib/reactions/iteration.js
devmanhinton: I personally like that style better, but I'm flexible and I'll change my style to your preference

yaymukund: Oops, should've been more specific. This style of assignment is a relic from semicolon-less JS code. Since we're adding semicolons, it should go too.

lib/reactions/iteration.js
devmanhinton: Awesome, good catch

lib/rules.js
```diff
@@ -59,16 +59,12 @@ rules.polymorphicValue = function(el) {
...
-  return el.children("." + name + "-").length
-}
-
-rules.collection = function(el) {
-  var collectionName = firstClassWhere(el, function(className) {
-    return this.hasSingularNamedChild(el, className)
-  }, this)
-  return collectionName ? undash(collectionName) : false
+rules.iteration = function(el) {
+  if(!(el.is("[data-each]"))) {
```
yaymukund: `return el.is('[data-each]')`

.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules
+bower_components
```
yaymukund, line 2: Why this change?

docs/iteration.md
```diff
@@ -0,0 +1,86 @@
...
+Backbone.Collection does not support attributes natively, but there are a number of
+options for extending collections to do so.  EndDash supports collection attributes
+as long as they are implemented as they are on Backbone.Model via the `get` method
+(which Backbone.Collection natively uses only for getting a model by id, not an
+attribute by name).  Typically collection attributes are used for metadata about
+the collection, such as total size (if the collection is paginated and this is
+different than length), as in the example below:
+
+```html
+<div class='users-' >
```
yaymukund, line 75: Doublequotes for HTML attributes?

.gitignore
devmanhinton, line 2: @bglusman

lib/reactions/model.js
```diff
@@ -8,9 +8,14 @@ var ModelReaction = Reaction.extend({
 
   init: function(next, reload) {
     var newModel = this.get(this.modelName);
-    if(!(newModel instanceof Backbone.Model || newModel instanceof Backbone.Collection ||
-          this.isEndDashCompatible(newModel))) {
-      newModel = new Backbone.Model(newModel || {});
+
+    if(!(newModel instanceof Backbone.Model || newModel instanceof Backbone.Collection
```
yaymukund: Do you need the first two checks anymore?

test/scope.js
```diff
@@ -0,0 +1,40 @@
+require('./support/helper');
+
+var expect = require("expect.js")
+  , fs = require("fs")
+  , generateTemplate = require("./support/generate_template")
+  , Backbone = require('backbone')
+  , variableReaction = require('../lib/reactions/variable');
+
+  describe ("With a template that has a polymorphic key but no scope changes", function(){
```
tobowers: would this be better as (pseudo code)

```
describe "scope", ()->
  describe "polymorphic key but no scope changes", ()->
    var markup = "blah"
    describe "model scoping reactions", ()->
      beforeEach ()->
        setup code from inside it

      it "should leave user", ()->

      it "should leave admin, ()->

```

test/support/templates/polymorphic.html
```diff
@@ -1,4 +1,6 @@
-<ul class = "things- typePolymorphic-">
-  <li class = "thing- whenAwesome-">awesome</li>
-  <li class = "thing- whenCool-">cool</li>
+<ul class = "things-">
```
tobowers: spacing consistency for class=""

test/iteration.js
```diff
@@ -0,0 +1,169 @@
...
+      this.things = new Backbone.Collection([
+        new Backbone.Model({ type: "awesome" }),
+        new Backbone.Model({ type: "cool" })
+      ]);
+
+      this.markup = fs.readFileSync(__dirname + "/support/templates/polymorphic.html").toString()
+      this.template = generateTemplate({ things: this.things }, this.markup)
+    })
+
+    it("should change the item when the type changes", function() {
```
tobowers, line 21: I think I wrote this while fixing a bug, but it ain't good... waaaaayyyyy too long of a test.

lib/reactions/model.js
devmanhinton: Still need 

``` javascript
instanceof Backbone.Collection
```

But not

``` js
instanceof Backbone.Model
```

Refactoring to have checks be in function 

``` js
isEndDashCompatible
```

test/support/templates/polymorphic.html
devmanhinton: Addressing in next commit, thanks!

test/iteration.js
devmanhinton, line 21: :) Cut down that spec quite a bit, INC (in next commit)

test/scope.js
devmanhinton: It is a bit clearer, INC

docs/iteration.md
devmanhinton, line 75: Punting this to the docs branch. Ie., when merge with that branch this will be overwritten

lib/rules.js
devmanhinton: Addressed in commit

devmanhinton:
@yaymukund  To address change from collection to iteration. It is a more semantic name. What that reaction does is iterate over the object in scope using its each method (and passing each to a template).

.gitignore
bglusman, line 2: old version of bower installs to components directory, new one to bower_components looks like, so if you resintall with bower it's all in bower_componenets now and not git ignored, hence showing up as uncommitted files all the time.

yaymukund:
@devmanhinton Not arguing that, but it makes it very difficult to review it. I'm just suggesting you send me a diff of just what has changed. If you can't do that, I can just review the whole file but it will take longer and you might have to fix stuff that you barely touched.

bglusman:
@devmanhinton if we revert the name change and make that a separate PR after this is merged it will be cleaner... the diffs will reflect actual difference...  I should have insisted on that when we were working, even having the name change be a separate commit might do it...

.gitignore
yaymukund, line 2: Wait, end-dash has a bower dependency? Where are these uncommitted files? How do they get there?

test/support/templates/collectionception.html
```diff
@@ -1,10 +1,10 @@
-<ul class = "things-">
-  <li class = "thing-">
-    <ul class = "items-">
-      <li class = "item-">
-        <ul class = "objects-">
-          <li class = "object-">
-            <div class = "name-"></div>
+<ul class="things-">
```
yaymukund, line 1: Sigh, maybe not for this PR but we should use a real life example like `worlds -> continents -> countries -> states -> cities -> districts`... this test is cute but nonsensical.

devmanhinton, line 1: Sure, we can do that but like you say, not part of this PR

devmanhinton:
@yaymukund Just did in the last 2 min

yaymukund:
Will this work with regular arrays or do they have to be backbone collections?

devmanhinton:
@yaymukund Yes & No. 

Depends as well if array has Backbone Models or literal models.

In general, EndDash will interpolate values of any model passed in (regardless of if it has a get method) but will only update on model/collection changes if that model/collection has get, set, on, & listenTo like Backbone objects.

So, in this case, with an array, it would interpolate the array's models into the DOM but would not update if an item was removed or added from, to, the array.

bglusman:
@yaymukund @devmanhinton maybe I misunderstand your response and/or mukund's question, but I thought he meant the iteration working with them?  Internally we convert arrays into backbone collections if not already one, that's one of the changes we made to get this to work I think.

Oh, but good point on not updating if items added to the array, if it was created from an array...  wonder if that's a real issue we need to worry about or not...

devmanhinton:
Ahhhh. Yes. Need to test again but yes. For the reason Brian mentioned (we wrap it in a Backbone.Collection)

docs/iteration.md
```diff
@@ -0,0 +1,86 @@
...
+
+var usersCollection = new Backbone.Collection([user1, user2, user3]);
+```
+
+Iterate through the collection using data-each.
+(You must be in scope of the collection)
+
+```html
+<div data-each>
+  <div>
```
yaymukund, line 19: Why do you need this inner div?

devmanhinton, line 19: Anything inside data-each is converted to a template (partial reaction) which runs through parser.js. In parser.js we have a check to make sure each template has only 1 root. Thus, it must have a single root.

This can be changed but I would punt that to a another PR once this is merged.

lib/reactions/collection.js
```diff
@@ -1,135 +1,116 @@
...
   addAll: function(models, index) {
-    index = index || 0
+    index = index || 0;
     if(_.isArray(models)) {
-      models = _(models)
-    }
-    models.each(function(item, index) {
-      this.add(item, index)
-    }, this)
+      models = _(models);
```
yaymukund, line 13: This is the wrong place to wrap `models`. Right now, I think we're doing:

``` js
init: function(model) {
  // We just store model blindly, no guarantees whether it's
  // backbone or non-backbone, underscore or
  // non-underscore.
  this.model = model;
},

addAll: function() {
  // Every time we call addAll, we have to do the dance where
  // we convert stuff into underscored stuff.
  if (_.isArray(model)) {
    model = _(model);
  }
},

// This isn't a great way to do it. At some point, maybe we have
// another method that works with the model as a collection. We'd
// have to do this whole dance again!
addSlightlyDifferent: function() {
  if (_.isArray(model)) {
    model = _(model);
  }
}
```

Instead, we should be doing that conversion once when we store the model. All other functions can just assume that it's been underscore-wrapped correctly.

Personally, I'd rather store it as a bare object and use underscore with the `_.each(models, function(model) {...})` syntax instead.

docs/iteration.md
yaymukund, line 19: I don't mind saving this for a subsequent PR, but this limitation should be documented so people aren't confused.

lib/reactions/collection.js
```diff
@@ -1,135 +1,116 @@
...
-      , template = new Template(model, { stack: this.stack.slice(0) })
-      , children
+    var Template = this.getTemplateClass(model),
+        template = new Template(model, { stack: this.stack.slice(0) }),
+        children;
 
-    template.template.data("model", model)
+    template.el.data("model", model);
 
     if(typeof index === "undefined") {
```
yaymukund, line 34: This is pretty complex, low level logic here. Could it be in a `util` class?:

``` js
util.insertAt(this.templates, template, index);
util.insertElementAt(this.el, template.el, index);
```

docs/iteration.md
devmanhinton, line 19: agreed

lib/reactions/collection.js
devmanhinton, line 34: could be, but that is a different PR, since this part of the code did not change in this PR.

lib/reactions/collection.js
```diff
@@ -1,135 +1,116 @@
...
-  remove: function(index, model) {
+  remove: function(model) {
     var templateToRemove = _(this.templates).find(function(template) {
-      return template.template.data('model') === model;
-    })
-    //if (templateToRemove) {
-      templateToRemove.template.remove()
+      return template.el.data('model') === model;
+    });
+      templateToRemove.template.remove();
```
yaymukund, line 51: `.el.remove()`?

lib/reactions/collection.js
devmanhinton, line 13: So,

``` js
  if (_.isArray(model)) {
    model = _(model);
  }
```

Is only hit when a collection is passed in a that top level without scoping. (otherwise it is wrapped by model.js)

I think a good solution to make this clearer is to wrap it in a Backbone.Collection in Template.js in bind.

If any of the above is unclear, feel free to ping me on campfire or in person

devmanhinton, line 13: See comment

```
//Hit only if array is passed in and iterated without scoping
```

lib/reactions/collection.js
devmanhinton, line 51: yes. We did not go through and get ride of template.template references. That was not the goal of this PR.

Happy to change and push that though or do you want to save all those changes for another PR?

test/iteration.js
yaymukund, line 21: Is this resolved? It still looks pretty long.

lib/reactions/collection.js
yaymukund, line 51: If you're already touching a line, it's generally a good idea to also clean up stuff in the process because... why not?

If you want to save it for another PR, that's fine too. Whatever you think is best.

devmanhinton, line 51: I'll just make it another PR this time. But in the future I'll try to make incremental changes as I find them. Was just so focused on the code when writing was not thinking about this unrelated, albeit necessary, change.

lib/reactions/model.js
```diff
@@ -8,9 +8,13 @@ var ModelReaction = Reaction.extend({
 
   init: function(next, reload) {
     var newModel = this.get(this.modelName);
```
yaymukund, line 10: `this.get(this.modelName) || {};`

Again, we have this branching `if (_.isArray(foo)) foo = new Backbone.Foo(foo);` stuff. That should be handled in one place.

devmanhinton, line 10: Please see the third discussion above this one.

Also, a lot of these are parts of the code that were the behavior of EndDash before this PR was started. I think that these are items are valid to consider and change but should be noted down as further PR.

I now have:

1) PR to replace template.template w/ template.el
2) PR to change when models are wrapped

Both of the above are valid and quick changes.

yaymukund, line 10: Isn't the `new Backbone.Collection` stuff new to this branch?

lib/reactions/model.js
```diff
@@ -22,6 +26,9 @@ var ModelReaction = Reaction.extend({
     if (!model) {
       return false;
     }
+    if (model instanceof Backbone.Collection) {
```
yaymukund, line 29: Won't this fail for the same reason that `instanceof Backbone.Model` failed? (i.e. different instance of `Backbone` used by the app than by EndDash internally)

devmanhinton, line 29: Yes, that is a good point. Perhaps define a helper that checks to see if .each is defined on the object?

lib/reactions/model.js
devmanhinton, line 10: Not totally, we did change it since it was missing an edge case with new collection iteration. So we changed it, slightly.

lib/reactions/model.js
devmanhinton, line 29: How would you feel about not wrapping objects at all. Assume that if you pass in a model that has the correct interface. And if it is a collection that it can be iterated through?

yaymukund, line 29: I feel like the correct solution is to use an existing backbone if one is defined when EndDash starts up.

I don't think that belongs in this PR, though, so don't worry about it. I'll open up a new issue.

devmanhinton, line 29: Great. Yes. It will be nice once that issue is resolved :D :dancer:

.gitignore
yaymukund, line 2: Bump!

test/iteration.js
devmanhinton, line 21: I believe the test that was 'too long' was 'should change the item when the type changes'. This was 10 lines. Now it is 6..

.gitignore
bglusman, line 2: run bower install in end-dash repo and see for yourself?  that's what the bower.json file is for, that's where inflection library comes from, and jquery and underscore.

bglusman, line 2: PS - doing Bump without mentioning anyone by name makes it harder to notice :-)

bglusman, line 2: which, taking my own advice, means I should @yaymukund you in a reply :-)

PR #57, opened 2013-11-01
Docs branch

bglusman:
@yaymukund @devmanhinton not perfect but hopefully a step forward.


docs/enddash-syntax.md
```diff
@@ -11,8 +12,9 @@
 ```js
 template.bind({
   user: new Backbone.Model({
-    firstName: 'Michael',
-    lastName: 'Jackson'
+    firstName: 'Tony',
+    lastName: 'Stark',
+    alias: 'IronMan'
```
yaymukund: Oh, hm, why change these? I guess it's not a big deal...

Can this be `'Iron Man'` instead of `'IronMan'`?

docs/enddash-syntax.md
```diff
@@ -21,27 +23,35 @@ template.bind({
...
-  addresses: {
+var mansion = new Backbone.Model({
+  resident: 'Tony',
+  placeList: {
     home: {
-      nickname: 'Graceland',
-      line1: '3717 Elvis Presley Blvd.',
-      line2: 'Memphis, TN 38116'
+      home: 'Stark Mansion',
+      areaNickname: 'Point Dume',
```
yaymukund: How about just `nickname`?

docs/enddash-syntax.md
```diff
@@ -21,27 +23,35 @@ template.bind({
...
 </div>
 ```
 
 ```js
-var elvis = new Backbone.Model({
-  nickname: 'Elvis',
-  addresses: {
+var mansion = new Backbone.Model({
+  resident: 'Tony',
+  placeList: {
```
yaymukund, line 39: `places` instead of `placeList`?

docs/enddash-syntax.md
```diff
@@ -21,27 +23,35 @@ template.bind({
...
+      home: 'Stark Mansion',
+      areaNickname: 'Point Dume',
+      city: 'Malibu',
+      state: 'CA'
     }
   }
 });
 
-template.bind(elvis);
+template.bind(mansion);
```
yaymukund, line 49: `template.el` belongs in a general section ("Usage"?) because it isn't specific to the "Changing scope" feature.

docs/enddash-syntax.md
```diff
@@ -21,27 +23,35 @@ template.bind({
...
+```
+For this template and this model, `template.el` will output:
+
 ```
```
yaymukund: It would actually output:

``` html
<div class="user">
  <p>Hello, <span class="resident-">Tony</span>! Your home is listed as:</p>

  <p data-scope="/placeList" class="home-">
    <span class="home-">Stark Mansion</span>,
    which is in
    <span class="areaNickname-">Point Dume</span>,
    a neighborhood in <span class="city-">Malibu</span>, <span class="state-">CA</span>
  </p>
</div>
```

docs/iteration.md
```diff
@@ -0,0 +1,93 @@
...
+</div>
+```
+
+Inside data-each there must be a single root with a set of dom elements.
+This set will be bound once to each model.
+
+On the resulting page you would see:
+
+```
+Tony
```
yaymukund: Ditto (please show the actual HTML output instead of converting it to plaintext.

docs/variables-and-conditionals.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+Conditionals
+============
+
+And a ternary operator is available for presence handling via 'truthiness' for attributes
+that may be present, with or without a false condition:
+
+```html
+<div class="user- #{availability ? available : unavailable}">
```
yaymukund, line 29: The `#{}` stuff might warrant its own section... thoughts? I'm not sure what all features it supports..

docs/enddash-syntax.md
bglusman: I don't think polymorphic will like a multi-word switch, can test when I build examples. And just changed to have a consistent theme for the docs, Iron Man being an Amicus 'thing'

docs/enddash-syntax.md
bglusman, line 39: Yeah, this felt weird for me with it not being a collection, and the version we ship will probably still have inflections in it... should work, but was afraid it might confuse a bit, dunno.

docs/enddash-syntax.md
bglusman: True, but we weren't doing "HTML" output but "appearance to user output" elsewhere, trying to be consistent.

docs/enddash-syntax.md
yaymukund: OH. I just realize `alias` isn't even being used. Can you just remove it from this example?

docs/variables-and-conditionals.md
bglusman, line 29: Yeah, maybe, it seemed related conceptually enough to the conditional endDash statements, but when we combine it all into one file we can rejigger the sections I guess, dunno.

yaymukund, line 29: Okay, I'll express that a bit stronger: given that `#{}` is an entirely different syntax from the usual `class="foo-"`, I think it should be in its own section. It's also important to explain that it only works within the `class="#{}"` and not free text (e.g. `<div>#{willNotWork}</div>`).

(But yeah, feel free to put it off until later but make sure it ends up getting separated before we merge.)

bglusman, line 29: Not just class but any attribute, but that definitely need clarifying, good thought/catch.

docs/enddash-syntax.md
yaymukund: Your comment doesn't answer the most important question... are you going to make it html or not?

docs/enddash-syntax.md
bglusman, line 49: I think I addressed what you mean here, look in context at this after latest commit.

docs/iteration.md
```diff
@@ -0,0 +1,106 @@
...
+var person3 = new Backbone.Model({firstName: 'Pepper', lastName: 'Potts', alias: 'none' });
+
+var peopleCollection = new Backbone.Collection([person1, person2, person3]);
+```
+
+Iterate through the collection using data-each.
+(You must be in scope of the collection)
+
+```html
+<div data-each>
```
jameselkins, line 18: Can you scope the collection explicitlyhere, giving the full example like https://github.com/Amicus/end-dash/blob/docs_branch/docs/enddash-syntax.md#changing-scope?  Otherwise it is confusing on how to do that.

docs/iteration.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+Polymorphic Iteration
+=====================
+
+To iterate over a collection, passing each model to a
+different template, based on a model attribute, add '<modelAttribute>Polymorphic-'
+
+```html
+<div class='aliasPolymorphic-' data-each>
```
jameselkins: Can you rename alias to something else?  Here it is used in the context of superhero name, but alias also has other programmatic meanings, making this hard to parse semantically at first glance.  My suggestion: heroName or maybe nomDeGuerre.

And also maybe make it clearer we are in the proper scope?

docs/iteration.md
```diff
@@ -0,0 +1,106 @@
...
+as long as they are implemented as they are on Backbone.Model via the `get` method
+(which Backbone.Collection natively uses only for getting a model by id, not an
+attribute by name).  Typically collection attributes are used for metadata about
+the collection, such as total size (if the collection is paginated and this is
+different than length), as in the example below:
+
+```html
+<div class='people-' >
+  <p>
+    There are <span class='totalCount-'></span> people allowed in Tony's basement.
```
jameselkins, line 89: I think we should leave this bit out of the public release unless we give an example of how to implement collection attributes.  I don't think it would be too difficult to rip out the piece in base_collection_presenter that deals with this and display it here.

docs/iteration.md
yaymukund: This doesn't feel like a great use of `polymorphic`, since presumably only one `person` will have a specific `polymorphic` value (e.g. only one person will be `IronMan`...). How about `characterType` in `['superhero', 'supervillain', 'human']`?

docs/partials.md
```diff
@@ -0,0 +1,49 @@
...
+```html
+<tr class="userRow">
+  <td class="firstName-"></td>
+  <td class="lastName-"></td>
+  <td class="alias-"></td>
+</tr>
+```
+
+The data-replace attribute tells EndDash to substitute the partial's root element for this element.
+Without data-replace, EndDash will embed the root element beneat the partial's element and leave it.
```
jameselkins: I think you should move this bit up to before you start to use data-replace.  Also, beneath is spelled incorrectly (missing last h)

docs/variables-and-conditionals.md
```diff
@@ -0,0 +1,50 @@
+Variable Interpolation
+======================
+
+## Variables
+
+EndDash variables are rendered into HTML elements when for display to the user:
```
jameselkins: are you missing some words here?  (between when <--> for)

docs/variables-and-conditionals.md
```diff
@@ -0,0 +1,50 @@
...
+
+```html
+<div class="user-">
+  My name is <span class="firstName-"></span> <span class="lastName-"></span>.
+</div>
+```
+But they may also be used to interpolate class names, id's or other attributes as below:
```
jameselkins: ..as below using the #{variableName} syntax .  Note the absence of the trailing dash.

docs/variables-and-conditionals.md
```diff
@@ -0,0 +1,50 @@
...
+```html
+<div class="user-" id="#{alias}">
+  My name is <span class="firstName-"></span> <span class="lastName-"></span>.
+</div>
+```
+
+Conditionals
+============
+
+And a ternary operator is available for presence handling via 'truthiness' for attributes
```
jameselkins: No need for the word "And".

docs/iteration.md
bglusman, line 18: @devmanhinton wanted it this way because that's no longer required, if you pass in a collection as the top level object to template.  We're going to make working examples for these, so hopefully that will demonstrate this clearly in context.

docs/variables-and-conditionals.md
```diff
@@ -0,0 +1,50 @@
...
+<div class="user- #{availability ? available : unavailable}">
+  <p>
+    My schedule is very full. <span class="isAvailable-">I just have a few openings</span>
+  </p>
+</div>
+```
+
+The same truthiness controls conditional visibility EndDash class elements that start with `is` or `has`,
+and their boolean opposites `isNot` and `hasNot`, as above with `isAvailable-`.  EndDash will hide (via a
+`display:none` style attribute) any such element when it's named attribute is falsy (or hide when truthy in
```
jameselkins: its should not have an apostrophe.

docs/iteration.md
bglusman: good suggestion @yaymukund I like that, I was struggling with good examples for that, that's a better one...

docs/variables-and-conditionals.md
jameselkins: Also let's have 2 separate examples - one for the ternary operator and one for the is/has/isnot/hasnot.  Maybe make it clearer that is and has are _exactly_ the same as well as their opposites, and the reason for having them both.

docs/variables-and-conditionals.md
bglusman: Hmm, not missing words, but perhaps poor writing :-)  perhaps: `EndDash variables are rendered into the body of HTML elements, displaying their values as ordinary text.` better?

jameselkins: great!

docs/iteration.md
devmanhinton, line 18: Hmmm. I did say that. I am also flex though. If we want to add the scope in because it makes it clearer, I am good with that as well.

jameselkins, line 18: yeah, that code would disambiguate things.

docs/iteration.md
bglusman, line 89: Good idea @jameselkins , but feels like something for examples folder, not for the docs?  Seems too detailed/not related to explaining/documenting EndDash to do it inline here.

docs/iteration.md
bglusman, line 18: We have another example that has explicit scope, I kind of like having one of each now, but if you definitely think it's unclear (not just because it used to not work this way without explicit scope) I can add it.

docs/enddash-syntax.md
bglusman, line 39: I'm leaving this for now, but comment again if you feel strongly/have another suggestion.

docs/enddash-syntax.md
```diff
@@ -21,27 +22,45 @@ template.bind({
 
 ```html
 <div class="user">
-  <p>Hello, <span class="nickname-"></span>! You live at:</p>
+  <p>Hello, <span class="resident-"></span>! Your home is listed as:</p>
```
yaymukund, line 25: `residentName`?

docs/iteration.md
```diff
@@ -0,0 +1,100 @@
...
+<div class='characterTypePolymorphic-' data-each>
+  <div class='whenHero-'>
+    // Models with Model.get('characterType') == 'hero' will bind to HTML here.
+    <span class='firstName-'></span> says: Don't worry.  I'll probably save you.
+  </div>
+  <div class='whenVillain-'>
+    // Models with Model.get('characterType') == 'villain' will bind to HTML here.
+    <span class='firstName-'></span> says: I'm going to kill Iron Man!
+  </div>
+  <div class='whenCivillian-'>
```
yaymukund: typo: `Civilian`

docs/iteration.md
yaymukund, line 89: :+1: @jameselkins agree completely.

yaymukund:
I'm merging this, but we need to make this much better before Wednesday. I'm going to copy these CR comments into issues so we can keep track of them.

docs/variables-and-conditionals.md
```diff
@@ -0,0 +1,51 @@
...
+```js
+template.bind({
+  user: new Backbone.Model({
+    firstName: 'Tony',
+    lastName: 'Stark',
+    alias: 'IronMan'
+    availability: ['10am', '2pm']
+  });
+});
+```
```
ChrisVoxland, line 51: Consider adding a model and "The resulting HTML will be:" to the examples on this page for clarity and consistency.

PR #71, opened 2013-11-01
needed for global reference in examples

bglusman:
@yaymukund this is needed for your examples directory to work after the template/view store changes eliminated EndDash variable in enddash.js file.

### DIFF
--- a/lib/end-dash.js
+++ b/lib/end-dash.js
@@ -53,4 +53,6 @@ var reactions = [
 
 _.each(reactions, Parser.registerReaction);
 
+EndDash = module.exports
+
 


yaymukund:
Can you be more specific? How is it broken and how can I replicate it?

bglusman:
@yaymukund I think if you try and run examples in master without this commit, then run them with it, you'll see, it shows EndDash as being an undefined variable without this (for me, on Friday anyway), and works with it.

devmanhinton:
@bglusman What is the status on this? 

It seems like it could be a bad decision to initialize a global variable in this file.

bglusman:
It's not normally a global, but it seems to be acting that way in examples.
 @yaymukund did you verify/reproduce? I think it needs to be in for
testers, unless I'm wrong about how it's happening.

215-460-9585 - _cell_
_Twitter_: @amicushq

yaymukund:
Interesting... you're correct. We need to rethink the build process at some point, I think. I'm not a fan of requiring browserify.

PR #72, opened 2013-11-04
Undo semicolons for now

bglusman:
@devmanhinton this will just undo the master merged stuff I did Friday, until we have other merges done cleanly.

### DIFF
--- a/lib/reactions/attribute.js
+++ b/lib/reactions/attribute.js
@@ -1,69 +1,69 @@
-var Reaction = require("../reaction"),
-    util = require("../util"),
-    rules = require("../rules"),
-    _ = require("underscore"),
-    get = util.get,
-    trim = util.trim,
-    interpolatorRegex = /#\{([a-zA-Z0-9_\- ]+)(?:\?([a-zA-Z0-9_\- ]+))?(?:\:([a-zA-Z0-9_\- ]*))?}/g;
+var Reaction = require("../reaction")
+  , util = require("../util")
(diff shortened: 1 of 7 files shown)


yaymukund:
Can we rebase the other branches so they work with the semicolons?

bglusman:
@yaymukund possibly... started to write that I think it will be messier and more dangerous to fix that way, but maybe if we do it with the flag to resolve all conflicts in favor or the other branches?  will mean we need another pass to clean those semi colons left out, but may be best way to deal with this specific set of conflicts...  makes me a bit nervous though having git auto resolve conflicts, dunno if it's worth the risk of moved stuff getting messed up somehow.

yaymukund:
> makes me a bit nervous though having git auto resolve conflicts, dunno if it's worth the risk of moved stuff getting messed up somehow.

Yeah, it's not worth the risk.

Can't you just rebase and resolve conflicts manually? If you need help doing this, let me know and we can pair for a bit.

bglusman:
@yaymukund I could, I just don't think it's the safest or easiest path forward... I think merging these in and doing them while there were open (non trivial) PR's against master was a mistake, and an easy one to undo and approach again when master is stable and merged.   welcome opinions from @devmanhinton or @tobowers, or we can try it without this and see what it's like, but it already feels overly complex without this added to the pile of conflicts.

devmanhinton:
closing this in favor of #77

PR #74, opened 2013-11-04
Interpolated variables inputs

devmanhinton:
@xcoderzach Yo.

Brian and I just wrote this.

Adds inputs back in. 

Not perfect -- still references rules.js. + uses old syntax 

``` js
<input class='variableName-' / >
```

Reasoning was we need to be as easily compatible with brief as possible for now (and to get it through CR)

Next steps would be to:

1) Get ride of rules.js
2) Make input/ contentEditable binding their own sytnax


devmanhinton:
Added flexibility so you can do:

``` html
<input value='firstName-' />
```

And it will do its normal interpolate value and bind thing

xcoderzach:
```
<input value='firstName-' />
```

is kind of weird, no?

devmanhinton:
hahahaha. I thought it was kind of nice, since you are saying bind the value of the input to this value on my model?

bglusman:
The nice thing about this is it avoids having a class name end-dash that doesn't change scope...  but data-input="firstName" or something does seem cleaner.

Actually, a lot of the pain from EndDash changes comes from the use of classes, particularly since designers hook into them when they're there, but that should really be discouraged... any chance we should consider having all end-dash behavior driven by data-reaction-type attributes to avoid this problem?

@devmanhinton / @xcoderzach thoughts?  tempted to get Topper & Adrian's thoughts too but start with your reactions.

tobowers:
hmm - I _like_ that designers hook into the classes... if you're not going to do that then why not handlebars?

bglusman:
@tobowers really?  It seems like an unstable thing that changes for reasons independent of design/semantics, but curious for the flip side, why is it specifically good, as opposed to having semantic classes that are JUST semantic and not functional like EndDash's are?

tobowers:
@bglusman because it enforces semantics and has good semantic hooks. When does the container ever independently change from semantic meaning of a container?

I agree about properties being bad in class names but semantic containers are nice here. It also gives designers a way to style before the javascript hooks in.

bglusman:
@tobowers Yeah, I guess since scope changes are part of the semantic structure of the document that makes sense for designers to (be able to) hook into and the new interpolation (and perhaps this data-input or input value syntax) will deal with the properties, that makes sense to me.

bglusman:
Wow, impressed this is still mergeable... but closing the PR for now, will reopen when it's actually ready to consider merging.

xcoderzach:
Why not merge them, would you want to merge the interpolated-variables branch without the interpolated-variables-inputs?

bglusman:
@xcoderzach no def not, just holding off on this till next sprint and deciding how we're handling transitioning brief to use interpolation, and perhaps whether to try and create support for deprecation or move it over in one big push.

xcoderzach:
> devmanhinton wants to merge 5 commits into interpolated-variables  from interpolated-variables-inputs

It's not merging into master :-)

PR #75, opened 2013-11-04
some suggested edits for clarity

pbadger:
Just copy changes


devmanhinton:
@pbadger We may be removing this section altogether. Thanks for looking into this!

PR #79, opened 2013-11-04
docs revisions for current state of master

bglusman:
@yaymukund @devmanhinton I think we should avoid merging this until we need to, but any revisions can be made here, and perhaps some of the changes will be merged in even under new syntax.

### DIFF
--- a/docs/iteration.md
+++ b/docs/iteration.md
@@ -48,26 +48,30 @@ In this case, `role` is behaving as a polymorphic attribute.
 
 ```html
 <div class="rolePolymorphic-" data-each>
-  <p><span class="firstName-"></span> says:</p>
+  <span class="firstName-"></span>
 
-  <div class="whenHero-"><span class="firstName-"></span> says: Don't worry. I'll probably save you.</div>
-  <div class="whenVillain-"><span class="firstName-"></span> <span class="lastName-"></span> says: Worry.</div>
-  <div class="whenCivilian-"><span class="firstName-"></span> says: Get me outta here!</div>
+  <div class="whenHero-"> says: Don't worry. I'll probably save you.</div>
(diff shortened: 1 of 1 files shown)


docs/iteration.md
```diff
@@ -1,73 +1,73 @@
...
-    <span class='firstName-'></span> says: I'm going to kill Iron Man!
-  </div>
-  <div class='whenCivilian-'>
-    // Models with Model.get('characterType') == 'civilian' will bind to HTML here.
-    <span class='firstName-'></span> says: I'm just a civilian!
-  </div>
+<div class="rolePolymorphic-" data-each>
+  <p><span class="firstName-"></span> says:</p>
+
+  <div class="whenHero-"><span class="firstName-"></span> says: Don't worry. I'll probably save you.</div>
```
yaymukund: Please make this and the example below DRY... you don't need:

``` html
<span class="firstName-"></span> says:
<span class="firstName-"></span> says:
<span class="firstName-"></span> says:
```

bglusman: I sort of understand and sort of don't... I agree it's not DRY, but it's not the same for all of them, one has last name two don't, and without that it doesn't expose the details of the model being used for interpolation.  Iron Monger shouldn't have "Iron says:", he needs "Iron Monger says:", so his template is different.  Maybe I'm confused though, or maybe you just object to the noise and think it's clear enough without accessing the instance attributes during iteration?

bglusman: Oh, I see, you extracted out and I didn't notice...your extraction wouldn't work though even if we only had first name because it's outside any when block... or maybe it would work?  I'm not sure actually.

yaymukund: I feel like it should work... good thing to test!

PR #81, opened 2013-11-05
update good part of docs change

bglusman:
@yaymukund good part of old docs branch I hope.

### DIFF
--- a/docs/iteration.md
+++ b/docs/iteration.md
@@ -42,7 +42,8 @@ var characters = new Backbone.Collection([
 
 ## Polymorphic attributes
 
-If your objects have an enum (or [Enumerated type](http://en.wikipedia.org/wiki/Enumerated_type)) field, you can specify handling based on which type it is. This is best explained with an example.
+If your objects have an enum (or Enumerated type) field, you can specify handling based on which type it is.
+This is best explained with an example.
 
 In this case, `role` is behaving as a polymorphic attribute.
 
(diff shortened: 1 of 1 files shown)


bglusman:
@devmanhinton or you can merge this in :-)

docs/iteration.md
```diff
@@ -1,73 +1,77 @@
...
-```
-<div class='firstName-'>Tony</div>
-<div class='firstName-'>James</div>
-<div class='firstName-'>Pepper</div>
-```
+## Polymorphic attributes
 
-Polymorphic Iteration
-=====================
+If your objects have an enum (or [Enumerated type](http://en.wikipedia.org/wiki/Enumerated_type)) field, you can specify handling based on which type it is. This is best explained with an example.
```
yaymukund: I don't like this wikipedia link. Are we also gonna link to http://en.wikipedia.org/wiki/Polymorphism_(computer_science) ? Let's keep docs uncluttered.

bglusman: Yeah, you added the mention of 'enum', which seemed good to me, but I don't feel like 'enum' without any context like that is standard enough to parse without at least a moment of confusion... maybe enumerated type is clearer, or maybe I'm just slow and everyone else will get it, but can see your point, just felt like it needed some clarification.

yaymukund: So are you gonna change it or not? Either is fine, but IDK what your response means.

bglusman: I was asking for an opinion on clarity, but yes, I'll change it :-)

215-460-9585 - _cell_
_Twitter_: @amicushq

PR #84, opened 2013-11-05
Add deprecation

devmanhinton:
Adds in support for deprecated collection iteration syntax

``` html
<div class='users-'>
  <div class='user-'>
```

### DIFF
--- a/lib/reactions/deprecated_collection.js
+++ b/lib/reactions/deprecated_collection.js
@@ -44,7 +44,7 @@ var CollectionReaction = Reaction.extend({
     }
   },
 
-  remove: function(index, model) {
+  remove: function(model, index) {
     var templateToRemove = _(this.templates).find(function(template) {
       return template.el.data('model') === model;
     })
@@ -77,7 +77,7 @@ var CollectionReaction = Reaction.extend({
       if(this.polymorphicKey) {
         this.collectionPresenter.on("change:" + this.polymorphicKey, function(model) {
           var index = this.collectionPresenter.indexOf(model);
-          this.remove(index, model);
+          this.remove(model, index);
           this.add(model, index);
         }, this);
       }
@@ -91,7 +91,7 @@ var CollectionReaction = Reaction.extend({
       }, this);
 
       this.collectionPresenter.on("remove", function(model, collection, opts) {
-        this.remove(opts.index, model);
+        this.remove(model, opts.index);
       }, this);
     }
   }
--- a/lib/reactions/deprecated_collection.js
+++ b/lib/reactions/deprecated_collection.js
@@ -45,11 +45,13 @@ var CollectionReaction = Reaction.extend({
   },
 
   remove: function(model, index) {
-    var templateToRemove = _(this.templates).find(function(template) {
-      return template.el.data('model') === model;
-    })
+    var indexToRemove,
+        templateToRemove = _(this.templates).find(function(template, index) {
+          indexToRemove = index;
+          return template.el.data('model') === model;
+      });
       templateToRemove.el.remove()
-      this.templates.splice(_(this.templates).indexOf(templateToRemove), 1);
+      this.templates.splice(indexToRemove, 1);
   },
 
   init: function(next) {
--- a/lib/reactions/deprecated_collection.js
+++ b/lib/reactions/deprecated_collection.js
@@ -1,0 +1,132 @@
+var EndDash = require('../end-dash')
+  , Reaction = require("../reaction")
+  , inflection = require("inflection")
+  , _ = require("underscore")
+  , get = require("../util").get
+  , rules = require("../rules");
+
+var CollectionReaction = Reaction.extend({
+
+  addAll: function(models, index) {
+    index = index || 0
+    if(_.isArray(models)) {
+      models = _(models)
+    }
+    models.each(function(item, index) {
+      this.add(item, index)
+    }, this)
+  },
+
+  getTemplateClass: function(model) {
+    if (this.polymorphicKey) {
+      var value = get(model, this.polymorphicKey)
+      return this.itemTemplates[value]
+    } else {
+      return this.itemTemplates["default"]
+    }
+  },
+
+  add: function(model, index) {
+    var Template = this.getTemplateClass(model)
+      , template = new Template(model, { stack: this.stack.slice(0) })
+      , children
+
+    template.el.data("model", model)
+
+    if(typeof index === "undefined") {
+      this.templates.push(template)
+      return this.el.append(template.el)
+    } else if(index === 0) {
+      this.templates.unshift(template)
+      this.el.prepend(template.el)
+    } else {
+      this.templates.splice(index, 0, template)
+      children = this.el.children()
+      template.el.insertAfter(children[index - 1])
+    }
+  },
+
+  remove: function(index, model) {
+    var templateToRemove = _(this.templates).find(function(template) {
+      return template.el.data('model') === model;
+    })
+      templateToRemove.el.remove()
+      this.templates.splice(_(this.templates).indexOf(templateToRemove), 1);
+  },
+
+  init: function(next) {
+    this.collection = this.get(this.collectionName) || []
+    this.collectionPresenter = this.getPresenter(this.collection)
+
+    this.stack.push(this.collection)
+
+    this.templates = []
+    this.el.html("")
+    this.el.data("collection", this.collection)
+    this.addAll(this.collection)
+    next()
+  },
+
+  observe: function(next) {
+    this.change(this.collectionName, function(model, newCollection) {
+      if(this.collection != newCollection) {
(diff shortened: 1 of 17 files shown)


devmanhinton:
@bglusman

lib/reactions/deprecated_collection.js
```diff
@@ -0,0 +1,132 @@
...
+  , rules = require("../rules");
+
+var CollectionReaction = Reaction.extend({
+
+  addAll: function(models, index) {
+    index = index || 0
+    if(_.isArray(models)) {
+      models = _(models)
+    }
+    models.each(function(item, index) {
```
yaymukund: Why am I repeating myself??

``` js
models.each(this.add, this);
```

bglusman: tone please

lib/reactions/deprecated_collection.js
```diff
@@ -0,0 +1,130 @@
...
+      template.el.insertAfter(children[index - 1])
+    }
+  },
+
+  remove: function(index, model) {
+    var templateToRemove = _(this.templates).find(function(template) {
+      return template.el.data('model') === model;
+    })
+      templateToRemove.el.remove()
+      this.templates.splice(_(this.templates).indexOf(templateToRemove), 1);
```
yaymukund: Interesting. Where's the `index` argument used?

I feel like we should be able to do this and traverse `this.templates` only once.

devmanhinton: Not sure what your second comment means.

But this functionality is present in case a user iterates through a collection and then latter adds a model to the collection at a certain index.

``` js
collection.add(model, {at: 3});
```

lib/reactions/deprecated_collection.js
```diff
@@ -0,0 +1,130 @@
...
+      }
+    }, this);
+
+    if(typeof this.collectionPresenter.on === "function") {
+
+      if(this.polymorphicKey) {
+        this.collectionPresenter.on("change:" + this.polymorphicKey, function(model) {
+          var index = this.collectionPresenter.indexOf(model);
+          this.remove(index, model);
+          this.add(model, index);
```
yaymukund, line 83: The interface for `add` and `remove` should be made more consistent.

lib/reactions/deprecated_collection.js
yaymukund: My second comment, elaborated:

``` js
// Search through the array.
var templateToRemove = _.find(...);
// Search through the array again!
var index = _.indexOf(...);

// Instead, we should be able to do something like:
var index = ...; // Search through the array once.
var templateToRemove = this.templates[index]; // Don't need to search again!
```

devmanhinton: My mistake. I thought we were discussing add.

Yes. I will do as another PR after the merge. Noting it now.

devmanhinton: Only not in this PR since was not change in this PR and refactoring internals was not goal of this PR (though maybe we should have spend more time increasing quality of old code first)

yaymukund: You're building out deprecated behavior, not copying old code. I feel like this should be built as carefully as we'd build a new feature.

lib/reactions/deprecated_collection.js
```diff
@@ -0,0 +1,130 @@
...
+        }, this);
+      }
+      this.collectionPresenter.on("sort reset", function() {
+        this.el.html("");
```
yaymukund: `this.el.empty()`

lib/reactions/deprecated_collection.js
devmanhinton: I see your point even if I do not fully agree. This same code is in collection.js. So we should change both.

lib/rules.js
```diff
@@ -1,7 +1,31 @@
...
+  var collectionName = firstClassWhere(el, function(className) {
+    return this.hasSingularNamedChild(el, className)
+  }, this)
+  return collectionName ? undash(collectionName) : false
+}
+
+rules.deprecatedIterationCheck = function(el) { //Deprecated
+  iterationName = this.deprecatedCollection(el) || this.deprecatedCollection(el.parent())
+  if (iterationName) {
+    console.log("****ENDDASH-DEPRECATED-ITERATION-SYNTAX-WARNING****: collection:" + iterationName)
```
yaymukund: Please remove `console.log` until we have a good way to disable them in production.

bglusman: How about checking for localhost and/or port in window.location?

yaymukund: We'd want to probably add a logging flag that's turned off in production builds by default. Or maybe off by default everywhere, and the user can turn it on as they please.

In any case, I think it's enough work to warrant its own branch.

bglusman:
@yaymukund I think everything to be addressed in this branch is done. LMK if you disagree/have anything else.

lib/reactions/deprecated_collection.js
bglusman: This is to be created as a separate issue/PR for now, can link here from there and/or vice versa once created.

lib/rules.js
```diff
@@ -1,7 +1,31 @@
...
+
+rules.deprecatedCollection = function(el) { //Deprecated
+  var collectionName = firstClassWhere(el, function(className) {
+    return this.hasSingularNamedChild(el, className)
+  }, this)
+  return collectionName ? undash(collectionName) : false
+}
+
+rules.deprecatedIterationCheck = function(el) { //Deprecated
+  iterationName = this.deprecatedCollection(el) || this.deprecatedCollection(el.parent())
```
yaymukund: This can be

``` js
return this.deprecatedCollection(el) || this.deprecatedCollection(el.parent());
```

bglusman: well, now it can be, while the console.log is commented out, but that's just temporary until we decide how to handle logging this...  seems silly to change the structure of the function temporarily.  Also seems nice for boolean methods to return true or false rather than truthy or falsy.

lib/reactions/deprecated_collection.js
yaymukund: @bglusman Guess I wasn't clear enough. I think it belongs in this PR because it's new code you're adding for deprecated behavior.

lib/rules.js
yaymukund:
``` js
return !!(this.deprecatedCollection(el) || this.deprecatedCollection(el.parent()));
```

We won't get to logging until next sprint, will we?

Branches in general should not have "temporary" code that gets fixed in subsequent branches. We have a spotty record of returning to old code to clean up infrastructure, so who knows how long this will stick around?

bglusman: @yaymukund That's why I wanted to leave in logging if the window.location included the key phrases, but you wanted a more permanent solution... not having this will make it harder for anyone getting brief up to date.  I agreed to pull it out to do in a separate PR, but now if we're not going to get to it in this sprint, I say let's put in the good-enough-for-now solution now and open an issue to refactor a more general logging solution for another sprint.

lib/reactions/deprecated_collection.js
bglusman: it's not new code, it's moved code, with minor tweaks, we took care to change as little as possible to avoid introducing extra risks or delays while supporting the old code paths for brief until we can remove them entirely, hopefully fairly soon.  Not having this in blocks master being usable by brief;  therefore, it has been kept as small and as little new code as we could manage, but you're expanding the scope of it asking to have that behavior refactored as if it were 'new'.

devmanhinton:
@yaymukund Both comments were addressed

deprecated_collections only iterate once through `this.templates`

And rules.js::deprecatedCollection check now just returns a boolean check

lib/rules.js
```diff
@@ -1,7 +1,24 @@
...
+
+rules.deprecatedCollection = function(el) { //Deprecated
+  var collectionName = firstClassWhere(el, function(className) {
+    return this.hasSingularNamedChild(el, className)
+  }, this)
+  return collectionName ? undash(collectionName) : false
+}
+
+rules.deprecatedIterationCheck = function(el) { //Deprecated
+  return iterationName = !!(this.deprecatedCollection(el) || this.deprecatedCollection(el.parent()));
```
yaymukund, line 19: Can you also not assign this to a variable?

bglusman:
I still don't like the lack of logging for developers hitting deprecated methods, that's not deprecated, that's just two ways of doing things, but I'm tired and don't want to fight about every little thing.  Maybe I'll start working on a logging PR tomorrow, I dunno.

yaymukund:
@bglusman Are the deprecated methods documented at all? If we don't document them, and we comment that they're deprecated (as we have), devs have no reason to be using them..

bglusman:
@yaymukund true-ish for external devs, but amicus devs have been using them, and that's who I'm concerned about.

PR #86, opened 2013-11-05
Consistent wrapping

devmanhinton:
@bglusman @yaymukund 

CODE REVIEW--------------------------------- no PULL REQUEST YET

Moved wrapping of objects to only be in parser.js && reactions/model.js
Moved wrapping code -> util.js

Question: Is it worth creating an EndDash stack class? This would deal with all the wrapping etc...

I could just extend the array prototype?

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -67,6 +67,32 @@ function findDescendantsAndSelf(element, selector) {
   return matches;
 }
 
+function wrapObject(model){
+  var newModel = model;
+  if(!(isEndDashCompatible(model))) {
+    if (_.isArray(model)){
+      newModel = new Backbone.Collection(model)
+    } else {
+      newModel = new Backbone.Model(model || {})
+    }
+  }
+  return newModel;
+}
+
+function isEndDashCompatible(model) {
+  if (!model) {
+    return false;
+  }
+  if (model instanceof Backbone.Collection) {
+    return true;
+  }
+  return (typeof model.on == 'function' &&
+          typeof model.once == 'function' &&
+          typeof model.set == 'function' &&
+          typeof model.get == 'function'
+         );
+}
+
 function get(obj, key) {
   var val;
   if(!obj) return;
@@ -82,5 +108,6 @@ module.exports = {
   getSelector: getSelector,
   get: get,
   trim: trim,
+  wrapObject: wrapObject,
   jqts: jqts
 };
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,4 +1,5 @@
-var _ = require("underscore");
+var _        = require("underscore"),
+    Backbone = require('backbone');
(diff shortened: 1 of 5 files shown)


yaymukund:
This looks close to being ready to merge. Good work. I'll leave some comments inline, but to answer your questions:

> Question: Is it worth creating an EndDash stack class? This would deal with all the wrapping etc...

Nope, I don't see the benefit and it seems like a lot of work.

> I could just extend the array prototype?

Definitely not. That's very invasive.

lib/util.js
```diff
@@ -67,6 +68,32 @@ function findDescendantsAndSelf(element, selector) {
   return matches
 }
 
+function wrapObject(model){
```
yaymukund, line 71: 1. Use `var foo = function() { ... };` instead of `function foo() { ... }` to avoid confusion with variable hoisting.
2. Call this something more precise, like `util.toEndDashModel(model)`

lib/util.js
```diff
@@ -67,6 +68,32 @@ function findDescendantsAndSelf(element, selector) {
...
+  var newModel = model;
```
yaymukund, line 72: Uhhh.

``` js
var toEndDashModel = function(model) {
  if (isEndDashModel(model)) {
    return model;
  }

  if (_.isArray(model)) {
    return new Backbone.Collection(model);
  } else {
    return new Backbone.Model(model);
  }
};
```

Please also be mindful of edgecases. We want it to behave predictably when the input is an empty object, `null`, or an invalid value (e.g. a `Number`).

lib/util.js
```diff
@@ -67,6 +68,32 @@ function findDescendantsAndSelf(element, selector) {
...
+    }
+  }
+  return newModel;
+}
+
+function isEndDashCompatible(model) {
+  if (!model) {
+    return false;
+  }
+  if (model instanceof Backbone.Collection) {
```
yaymukund, line 87: Do we need to check both `instanceof Backbone.Collection` **and** `model.on && model.once && etc.`? Don't `Backbone.Collection`s already implement the `on`/`once`/`set`/`get` interface?

devmanhinton, line 87: Good thinking. I think I will just remove that check and just see if it is EndDash compatible instead. Good call Mukund.

lib/util.js
devmanhinton, line 72: Yes. Good point.

lib/util.js
devmanhinton, line 71: Okay. Was looking for a better name, thanks.

I followed file convention. I'll change the whole file to `var foo = function() { ... }` and then semicolonize the file.

devmanhinton:
@yaymukund  Thanks for feedback. Making changes now + after lunch.

PR #87, opened 2013-11-05
Readme notes

bglusman:
@yaymukund @sethbannon readme note for reviewers on pending change

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Include the library and dependencies:
 
 ## Templating from models
 
-Define your templates<sup>see note below</sup>:
+Define your templates:
 ```html
 <script type="text/enddash" name="character">
   <div class="user">
--- a/README.md
+++ b/README.md
@@ -10,19 +10,29 @@ Include the library and dependencies:
 <script src="/scripts/end-dash.js"></script>
 ```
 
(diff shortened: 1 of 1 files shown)


yaymukund:
I don't think we need to add this giant disclaimer.

a) We'll do a version bump when we make this change, so it's not like apps will break.
b) It's in beta, so they should expect changes to the public interface. It's one of the acknowledged tradeoffs to trying out shiny new things.

bglusman:
cc'ing seth via email since he may not follow github discussions...

I think we can and should remove it as soon as it's merged in, and perhaps
modify the language, but it's a big enough change and there's a viable
option for them that anyone who happens to be playing with it for now
should know about it.  It's temporary, but I think kind of important to
someone providing feedback or trying it out.

215-460-9585 - _cell_
_Twitter_: @amicushq

devmanhinton:
I actually agree with Mukund on this. Brian, I think it is great to think of, and try to make these beta developers feel good, but since it is labeled "beta" such extra note impending change is unnecessary :D

sethbannon:
I do follow github discussions :eyes: 

I'll defer to the wisdom of you guys on this one.

PR #88, opened 2013-11-05
Change util.js to have use var name = function syntax

devmanhinton:
@yaymukund 

Refactor to use 

``` js
var name = function() {

}
```


devmanhinton:
nvm, non-most recent master... weird. Deleting and then repushing

yaymukund:
how does this have conflicts so fast? Please fix and I can merge. BTW, you have run tests and tested it in brief, right?

yaymukund:
D'oh, you beat me to it.

PR #89, opened 2013-11-05
Refactor util.js to put all its exports on the export object

devmanhinton:
@yaymukund Refactor util.js to export functions on the export object :)

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,12 +1,11 @@
 var _ = require("underscore");
+
 /**
- * Returns a unique selector from the root element
- * to el
- *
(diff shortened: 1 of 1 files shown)


lib/util.js
```diff
@@ -42,21 +41,21 @@ function getSelector(el, root) {
...
 
   return selector;
 }
 
-function trim(str) {
+exports.trim = function(str) {
   return (str || "").replace(/^\s+|\s+$/g, "");
 }
 
-function jqts(element) {
```
yaymukund, line 55: This is no longer used anywhere, I don't think.

PR #91, opened 2013-11-05
Refactor to util

devmanhinton:
This refactors the wrapping of EndDash objects to be done from Template.js and Model.js
via util.js. @bglusman @yaymukund

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -91,14 +91,13 @@ var isEndDashCompatible = function(model) {
          );
 }
 
-exports.toEndDashObject = function(model) {
-  var newModel = model;
-  if(!(isEndDashCompatible(newModel))) {
-    if (_.isArray(newModel)){
-      newModel = new Backbone.Collection(newModel);
-    } else {
-      newModel = new Backbone.Model(newModel || {});
-    }
+exports.toEndDashCompatibleModel = function(model) {
+  if (isEndDashCompatible(model)) {
+    return model;
   }
-  return newModel;
+  if (_.isArray(model)){
+    return new Backbone.Collection(model);
+  } else {
+    return new Backbone.Model(model || {});
+  }
 }
--- a/lib/util.js
+++ b/lib/util.js
@@ -75,3 +75,30 @@ exports.get = function(obj, key) {
     return obj[key];
   }
 }
+
+var isEndDashCompatible = function(model) {
+  if (!model) {
+    return false;
+  }
+  if (typeof model !== 'object') {
+    throw new Error('Tried to bind template to a ' + typeof model +
+                    ', but templates can only be bound to objects.');
+  }
+  return (typeof model.on == 'function' &&
+          typeof model.once == 'function' &&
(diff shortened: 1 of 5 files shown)


lib/util.js
```diff
@@ -74,4 +74,31 @@ exports.get = function(obj, key) {
   } else {
     return obj[key];
   }
+}
+
+var isEndDashCompatible = function(model) {
```
yaymukund, line 79: Could this be `isEndDashObject`? Seems more natural.

lib/util.js
```diff
@@ -74,4 +74,31 @@ exports.get = function(obj, key) {
...
+                    ', but templates can only be bound to objects.');
+  }
+  return (typeof model.on == 'function' &&
+          typeof model.once == 'function' &&
+          typeof model.set == 'function' &&
+          typeof model.get == 'function'
+         );
+}
+
+exports.toEndDashObject = function(model) {
```
yaymukund: See my old comment here: https://github.com/Amicus/end-dash/pull/86#discussion_r7438976

lib/util.js
bglusman, line 79: disagree, there is no such thing as an "EndDashObject", there's an EndDash duck type... I guess isEndDashType could convey that as well or better than 'compatible' does, but don't think 'object' does.

bglusman, line 79: oh, sorry, I seem to be confused, hadn't seen EndDashObject stuff above... not sure I like there, but still reading, sorry.

yaymukund, line 79: `toEndDashModel` was my original suggestion, but @devmanhinton went with `toEndDashObject`. I just want to keep the `to*` and `is*` methods consistent.. don't care too much what we call it. :/

devmanhinton, line 79: Changed to 'toEndDashCompatibleModel'. Issue resolved

lib/util.js
devmanhinton: That comment has no relevance to this line as it stands

yaymukund: Hm, I think it's clearer to have early termination there.

``` js
if (isEndDashCompatible(model)) {
  return model;
}

// do the actual conversion
```

Plus, you don't need to assign stuff to `newModel`. You can just return `new Backbone.Collection` directly.

Plus, there's absolutely no reason to assign `newModel = model`

**EDIT:** My old comment had example code that did all these things:

``` js
var toEndDashModel = function(model) {
  if (isEndDashModel(model)) {
    return model;
  } else if (_.isArray(model)) {
    return new Backbone.Collection(model);
  } else {
    return new Backbone.Model(model);
  }
};
```

devmanhinton: Ahhh. Cool. Yes. That looks better. (That was not in the other comment though :D)

lib/util.js
yaymukund, line 79: `isEndDashCompatible(model)` should now change to `isEndDashCompatibleModel(model)`

PR #92, opened 2013-11-05
Readme update

bglusman:
more consistent example with our docs... need to update examples next


bglusman:
@devmanhinton removed the note since you guys didn't like but want to get rid of that michael jackson thing since all our docs are iron man now.

PR #93, opened 2013-11-06
changed collections to only iterate once through this.templates when rem...

devmanhinton:
@yaymukund This changes collection.js to only iterate once through this.templates in #remove. 

You merged in code that did this for deprecated collections yesterday.


yaymukund:
We should remove that deprecated code asap in next sprint.. yuck!

PR #94, opened 2013-11-06
missing backbone require

bglusman:
@devmanhinton   this was missing, causes tests to fail when run individually, though they pass en masse.


devmanhinton:
Good call! Thank you for clarifying :D

PR #95, opened 2013-11-06
This is a start to cleaning up rules / collections / polymorphic iterati...

devmanhinton:
Starting to clean up rules.js & collection.js

As an aside,

instead of polymorphicIteration, what do you think of conditionalIteration? 

@bglusman

### DIFF
--- a/lib/rules.js
+++ b/lib/rules.js
@@ -58,29 +58,31 @@ rules.negationClass = function(className) {
   return !!className.match(/(?:^|\s)(?:isnt|isNot|hasNo)(\S+)-(?:$|\s)/)
 }
 
-rules.polymorphicKeyClass = function(className) {
-  var match = (className || "").match(/(\S+)Polymorphic-(?:\s|$)/)
-  return (match) ? match[1] : false
+rules.polymorphicKeyClass = function(el) {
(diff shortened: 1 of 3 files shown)


lib/rules.js
```diff
@@ -100,7 +102,7 @@ rules.variable = function(el) {
     return !this.viewClass(className) &&
            !this.conditionalClass(className)
   }, this)
-  if (this.polymorphicKeyClass(variableName) || this.polymorphicValueClass(variableName)) {
+  if (this.hasPolymoprhicClass(el)) {
```
yaymukund: Typo.

Please spend some time before a PR going over your own diff to prevent stuff like this. This isn't the first typo you've tried to merge.

devmanhinton: Lol.

Lets clarify. This is a spelling typo. That is consistent throughout. Thus the code still works fine. I simply misspelled a word, which from the computer's perspective makes no difference.

But yes, ideally it would be spelled correctly. I'll change it and push it up.

yaymukund: I see. Thank you for the clarification.

devmanhinton: sure :D :thumbsup:

PR #96, opened 2013-11-06
The parser does not use backbone events or backbone, not need to require...

devmanhinton:
Explanatory. No need to have backbone events or backbone in the parser

@bglusman

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -1,5 +1,4 @@
-var Backbone = require("backbone"),
-    Template = require("./template"),
+var Template = require("./template"),
     path = require('path'),
     util = require("./util"),
     _ = require("underscore"),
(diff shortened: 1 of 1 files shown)


bglusman:
Cool, happy to merge this but why does it currently extend events?  Was that usage moved somewhere else, template or template store? Just want to be sure it's not needed.

devmanhinton:
To be honest I am not 100% sure. I'm guessing it was moved to template_store. Or else Zach just extended it that way by default (either he was planning something that never came to fruition or else deleted some functionality).

Ran specs == all pass. I can't see any clear use anywhere in EndDash. Parser is only used to do

new Parser(args).generate().

:D

devmanhinton:
@bglusman  bump! :)

PR #99, opened 2013-11-06
Refactor polymorphic branching

devmanhinton:
@yaymukund @bglusman 

Good refactor of collection reactions. Specifically their parsing.

There is more code then before and longer variable names. Longer code = bad. But I'm writing in a trade-off. Less succinct code for much clearer code.

Please let me know what you think.

### DIFF
--- a/lib/reactions/collection.js
+++ b/lib/reactions/collection.js
@@ -16,7 +16,7 @@ var CollectionReaction = Reaction.extend({
       var value = get(model, this.polymorphicKey);
       return this.childrensClassTemplates[value];
     } else {
-      return this.childrensClassTemplates["nonPolymorphic"];
+      return this.childrensClassTemplates[this.nonPolymorphicEnding];
     }
   },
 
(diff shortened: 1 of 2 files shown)


lib/reactions/collection.js
```diff
@@ -88,25 +88,44 @@ var CollectionReaction = Reaction.extend({
   }
 }, {
   selector: "[data-each]",
-    parse: function(el, state) {
```
bglusman, line 91: I guess the indentation here was just messed up before?

yaymukund:
What's a `nonPolymorphicEnding`? How is this code better?

lib/reactions/collection.js
devmanhinton, line 91: Yes. It was.

devmanhinton:
@yaymukund 

This code is better because

-it clearly branches polymorphic and nonPolymorphic template parsing into two distinct parts
-it makes template lookUp clearer (before nonPolymorphic and Polymorphic named their templates
   lastItemOnPathStack + "default", when there was no whenValue.) nonPolymorphicEnding is the ending EndDash now adds onto templates if it is not polymorphic iteration
-Naming is better, templates is now childrensTemplates (which is a more accurate name)
-Better decomposition, related to first item, but before parse was doing several things without calling sub-functions which === bad decomp.

lib/reactions/collection.js
```diff
@@ -88,25 +88,44 @@ var CollectionReaction = Reaction.extend({
...
+  },
+
+  parseDefault: function($child, state) {
+    var name = _.last(state.pathStack) + this.nonPolymorphicEnding;
+    this.registerAndRemoveChild($child, name, this.nonPolymorphicEnding);
+  },
+
+  registerAndRemoveChild: function($child, name, condition) {
+    EndDash.registerTemplate(name, $child);
+    this.childrensTemplates[condition] = EndDash.getTemplateClass(name);
```
yaymukund: I think internally, we should be using methods on `EndDash.TemplateStore` directly so dependencies are clearer.

If it's too verbose, you can also alias `templateStore` at the top of the file:

``` js
var ...
    EndDash = require('../end-dash'),
    templateStore = EndDash.templateStore;
```

devmanhinton: No. I agree with you.

For now, I'd prefer it to verbose and clear.

Let me change this.

bglusman:
can't merge now, probably semi colins

yaymukund:
The more I read the polymorphic code, the more I feel like we should just call this `switch-case` instead of `polymorphic`. I feel like that would be a lot clearer. The fact that we're branching on an attribute of the model is only incidental— for example, we branch on `call.status` but that isn't really a polymorphic attribute.

Also, maybe we could move it to its own reaction. Is there any reason for `polymorphic`(a.k.a. `switch-case`) to be in the same reaction as the collection stuff?

tobowers:
:+1: on what mukund said... It's true that it's not necessarily polymorphic, but it is always a switch.  What about:

``` html
<div class="switchStatus- call-">
  <p class="whenHangup-">
    You done hungup
  </p>
</div>
```

bglusman:
yeah, I like calling it a switch in some version too, particularly now we support an `else` case (and making that into `elseWhen-` or `else-` or something instead of just no class would be nice)

yaymukund:
Errr, not sure I understand, @amicus-jenkins. Do you mean "Can one of the members of the end-dash team approve/merge this PR?"

While the old code wasn't great, I don't think this is much clearer either, to be honest.

bglusman:
please test is what jenkins is looking for mukund

On Thursday, November 14, 2013, Mukund Lakshman wrote:

> Errr, not sure I understand, @amicus-jenkinshttps://github.com/amicus-jenkins.
> Do you mean "Can one of the members of the end-dash team approve/merge this
> PR?"
> 
> While the old code wasn't great, I don't think this is much clearer
> either, to be honest.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/Amicus/end-dash/pull/99#issuecomment-28487879
> .

## 

215-460-9585 - _cell_
_Twitter_: @amicushq

yaymukund:
> please test is what jenkins is looking for mukund

Huh?

Am I supposed to type "please test?" Are you asking me to actually run the tests myself? Why? I don't think it should be merged even if it passes tests..

yaymukund:
Oh, it's an automated message by [this plugin](https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin)... yeah I'm going to ignore it.

devmanhinton:
lol. I am going to close this for now.

PR #100, opened 2013-11-06
Update existing examples

bglusman:
@devmanhinton

### DIFF
--- a/examples/changing-scope.html
+++ b/examples/changing-scope.html
@@ -44,9 +44,17 @@
         places: {
           home: {
             name: 'Stark Mansion',
-            area: 'Point Dume',
             city: 'Malibu',
-            state: 'CA'
+            state: 'CA',
+            area: 'Point Dume'
+          },
+          work: {
+            name: 'Stark Industries',
+            street: '200 Park Ave.',
+            city: 'New York City',
+            state: 'NY',
+            zip: '10023',
+            area: 'Manhattan'
           }
         }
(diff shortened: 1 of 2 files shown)


examples/changing-scope.html
```diff
@@ -19,13 +19,18 @@
...
-
-      <address data-scope="/addresses" class="home-">
-        <p class="nickname-"></p>
-        <p class="line1-"></p>
-        <p class="line2-"></p>
-      </address>
+      <p>Hello, <span class="resident-"></span>! Your home is listed as:</p>
+
+      <p class="places-">
+        <span class="home-">
```
devmanhinton, line 25: Is it weird to have 'places-' as the name since there is only once place ('home')?

bglusman, line 25: Maybe, I had it as 'placeList' before but Mukund didn't like that...  now that we're not using inflection (necessarily) models can be plural even if not collections, but his example was mostly to show multiple levels of model nesting.  The idea seems to be home is a place you have to have, but you may have other places that other templates might use... maybe can expand on this in example further, like hasWork- and hasGPS- but dunno.

devmanhinton, line 25: Sure! I would either add more one more place then?

Or else change it to 'place-'. 

Do not want people to think this is clunky :)

Otherwise, good to merge.

devmanhinton:
Cool, great.

PR #110, opened 2013-11-07
Document set presenter

bglusman:
@devmanhinton this is branched off of the setGetPresenter PR, since it documents what I added there.

### DIFF
--- a/docs/presenters.md
+++ b/docs/presenters.md
@@ -1,0 +1,44 @@
+Presenters
+===============
+
+If you wish to follow the Model-View-Presenter pattern, EndDash supports a hook
+to specify what presenter to use for a given model.  By default, this function
+simple returns the model itself, a simple identity function, but by passing in
+your own lookup function to `EndDash.setGetPresenter` EndDash will run your code
+instead to load a presenter if one is found.  Here's an example getPresenter
+function that looks for a custom presenter in config.presenterDirectory and uses
+a default base presenter or base collection presenter when no presenter specific
+to the model's name attribute is defined:
+
+```js
+  function getPresenter(model) {
+    var modelName = inflection.underscore(model.name || "")
+      , id = model.cid
+      , Presenter
+      , basePresenter
+
+    if(!(model instanceof Backbone.Model) && !(model instanceof Backbone.Collection))
+      return model
+    //give collections a unique id
+    if(model instanceof Backbone.Collection) {
+      if(!model.cid) id = model.cid = _.uniqueId("collection")
+      basePresenter = "/base_collection_presenter"
+    } else {
+      basePresenter = "/base_presenter"
+    }
+    if(presenters[id]) {
+      return presenters[id]
+    } else {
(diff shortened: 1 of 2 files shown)


lib/end-dash.js
```diff
@@ -19,6 +20,8 @@ exports.registerView = _.bind(viewStore.load, viewStore);
 
 exports.setCustomGetView = _.bind(viewStore.setCustomGetView, viewStore);
 
+exports.setGetPresenter = _.bind(Reaction.setGetPresenter, Reaction)
+
```
devmanhinton, line 24: Was this not in the other PR?

docs/presenters.md
```diff
@@ -0,0 +1,44 @@
...
+        if(e.code !== "MODULE_NOT_FOUND") {
+          throw e
+
+        }
+        Presenter = require(config.presenterDirectory + basePresenter)
+      }
+      return presenters[id] = new Presenter(model)
+    }
+  }
+```
```
devmanhinton: Do we need this big of a note / example. Since, for now, it is not a core part of EndDash but an add on. What about just document that this exists. This PR up to line 8?

lib/end-dash.js
bglusman, line 24: yeah, it was, shouldn't show up in the diff now, that seems weird that it is, thought maybe nothing happened to invalidate github's cached diff or something

docs/presenters.md
bglusman: Yeah, that's what I was mentioning before, questionable, but it felt a little naked just having a whole doc file to say "there's this method you can use", but it didn't fit in anywhere else, and thought example might help clarify intention.  Maybe see what someone else thinks, or just take out, I dunno

devmanhinton: I say just take it out. But will ping @jameselkins . When you get a chance, please let us know what you think. Thank you for your time doing this. :)

bglusman: Yeah I think you're right, I'll just remove it.  If you disagree or think a simpler example is justified let us know james

jameselkins: @bglusman I am confused by this doc. It says you have to pass "your own lookup function to EndDash.setGetPresenter".  However, I don't see anything passed to setGetPresenter.

Also, the concept of presenters should be explained with like 1-2 sentences.. when I started using EndDash I had no idea what presenters were/what they were used for.  Even now, we (as a team) sometimes debate about which logic should be in presenters vs views vs representers.. for this doc to be excellent it would be amazing to have guidance on that issue.

To address @devmanhinton 's implied question, I think that you should leave it in.  Why document a feature without showing people how to actually use it?

lib/end-dash.js
devmanhinton, line 24: This looks great to merge except for this weird diff thing. Any thoughts on this?

bglusman, line 24: nah, think it's fine, just a github bug I think.

PR #112, opened 2013-11-07
Handle looping edge cases

devmanhinton:
@bglusman Yo, yo, yo. EndDash looping now handles edgeCases.

Currently that just means a div in data-each that is unnamed

``` html
<div 'kindPolymorphic-' data-each>
  <div class='whenUser-'>
  </div>
  <div data-im-base-case*>
  </div>
</div>
```

*this is not necessary just for fun!

We can decided on a reserved word if we want. That is easy to do.

### DIFF
--- a/lib/reactions/looping.js
+++ b/lib/reactions/looping.js
@@ -14,7 +14,7 @@ var LoopingReaction = Reaction.extend({
   getTemplateClass: function(model) {
       var template = this.itemTemplates[get(model, this.polymorphicKey) || "default"];
       if (!template) {
-        template = this.itemTemplates["default"]
+        template = this.itemTemplates["default"];
       }
       return template;
   },
--- a/lib/reactions/looping.js
+++ b/lib/reactions/looping.js
@@ -12,12 +12,11 @@ var LoopingReaction = Reaction.extend({
   },
 
   getTemplateClass: function(model) {
-    if (this.polymorphicKey) {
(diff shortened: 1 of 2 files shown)


lib/reactions/looping.js
```diff
@@ -12,12 +12,11 @@ var LoopingReaction = Reaction.extend({
   },
 
   getTemplateClass: function(model) {
-    if (this.polymorphicKey) {
-      var value = get(model, this.polymorphicKey);
-      return this.itemTemplates[value];
-    } else {
-      return this.itemTemplates["default"];
-    }
+      var template = this.itemTemplates[get(model, this.polymorphicKey) || "default"];
```
bglusman: looks like this got indented one too far

devmanhinton: You are right. Thank you catching. It is fixed.

lib/reactions/looping.js
```diff
@@ -12,12 +12,11 @@ var LoopingReaction = Reaction.extend({
...
+      if (!template) {
+        template = this.itemTemplates["default"];
```
bglusman: It's a little confusing why there's the || "default" above and then THIS 'default' here....  shouldn't line 15 be doing exactly what this line is doing, if there is a "default" template, without this line being there?

devmanhinton:
@bglusman It should be clear now. It is a bit tricky code.

PR #114, opened 2013-11-08
add a checkbox to inputs

bglusman:
@devmanhinton

### DIFF
--- a/examples/inputs.html
+++ b/examples/inputs.html
@@ -33,14 +33,16 @@
 
   <script type="text/enddash" name="whichCharacter">
   <div>
-    <p>Who is your favorite character?</p>
-    <input type="radio" class="name-" name="name-" value="tony" id="tony"/>
+    <p>Love to hate?</p>
+    <input type="checkbox" class="disliked-"/>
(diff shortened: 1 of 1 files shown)


examples/inputs.html
```diff
@@ -53,12 +55,12 @@
...
 
-      var favoriteCharacter = new Backbone.Model({
-        name: 'tony'
+      var extremeCharacter = new Backbone.Model({
+        name: 'Tony'
       });
 
       var template = EndDash.getTemplate('inputName', you),
-          template2 = EndDash.getTemplate('whichCharacter', favoriteCharacter);
+          template2 = EndDash.getTemplate('whichCharacter', extremeCharacter);
```
devmanhinton, line 63: Should we make this two separate templates (inputs vs. checkboxes.) Do you think someone looking at EndDash for the first time would understand it better, if they were two examples rather then one?

bglusman, line 63: <shrug> I dunno... I'd say we can do that later if desired, but don't really think it's necessary just for a single checkbox :-)

devmanhinton, line 63: Alright... :D

PR #115, opened 2013-11-08
add minimal debugger docs, fix typo in file.

bglusman:
@devmanhinton minimal, maybe we can do better?  worth having an example for?

### DIFF
--- a/docs/debugger.md
+++ b/docs/debugger.md
@@ -1,0 +1,24 @@
+Debugger
+======
+
+## Debugging
+
+Open up a debugger in development mode to inspect the context of the
+template.
+
+Open a context in the highest template scope:
+```html
+<body>
+  <div debugger />
+</body>
+```
+
+In a child model scope:
+```html
+<body>
+  <div class="questions-">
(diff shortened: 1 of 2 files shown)


docs/debugger.md
```diff
@@ -0,0 +1,24 @@
...
+```html
+<body>
+  <div debugger />
+</body>
+```
+
+In a child model scope:
+```html
+<body>
+  <div class="questions-">
```
devmanhinton, line 19: This looks great! Do you think it is clear enough we are assuming a model has been passed in a with a property 'questions-' which is a object?

bglusman, line 19: fair question...  guess it can't hurt to add that in above, though it seems pretty clear to me in context of the overall project.

devmanhinton, line 19: Did you not like the data-scope that was on there before?

bglusman, line 19: Well, it was wrong, because it was data-scope='questions-' rather than 'questions' so I opted for just using class :-)

PR #118, opened 2013-11-08
Scoping docs

devmanhinton:
@bglusman

### DIFF
--- a/docs/scoping.md
+++ b/docs/scoping.md
@@ -54,7 +54,7 @@ Normal UNIX path shorthands apply: `..` to move back up a scope level, `/` to se
   //User scope
   <div class='hobby-'>
   //Hobby scope
-    <div data-scope='..'>
+    <div data-scope='../'>
     //Back in User Scope
       <div data-scope='/user/hobby'>
       //Back in Hobby scope
--- a/docs/scoping.md
+++ b/docs/scoping.md
@@ -64,6 +64,6 @@ Normal UNIX path shorthands apply: `..` to move back up a scope level, `/` to se
 </div>
 ```
 
-'`class='user-'` is actually syntatic sugar for `data-scope='./user-'`.
+'`class='user-'` is actually syntatic sugar for `data-scope='./user'`.
 
 
--- a/docs/scoping.md
+++ b/docs/scoping.md
@@ -54,9 +54,9 @@ Normal UNIX path shorthand apply: `..` to move back up a scope level, `/` to sep
   //User scope
   <div class='hobby-'>
   //Hobby scope
-    <div class='..'>
+    <div data-scope='..'>
     //Back in User Scope
-      <div class='/user/hobby'>
+      <div data-scope='/user/hobby'>
       //Back in Hobby scope
       </div>
     </div>
--- a/docs/scoping.md
+++ b/docs/scoping.md
@@ -1,9 +1,9 @@
 What is Scoping?
 ================
 
-Scope in EndDash refers to the model that is on top of the EndDash stack.
-Every template and partial has its own scope. The `root` scope is always the object passed
-to EndDash's 'bind' or 'getTemplate'.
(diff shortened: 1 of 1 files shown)


docs/scoping.md
```diff
@@ -0,0 +1,67 @@
+What is Scoping?
+================
+
+Scope in EndDash refers to the model that is on top of the EndDash stack.
+Every template and partial has its own scope. The `root` scope is always the object passed
```
bglusman: should `root` be in backticks like this?  we don't use it was a keyword/reserved word or anything, I think untick it or they may think they can use root to refer to it or somehting.

docs/scoping.md
```diff
@@ -0,0 +1,67 @@
...
+to EndDash's `bind` or `getTemplate`.
```
bglusman: Are we 'advertising' bind as part of public interface?  Thought we were only promoting getTemplate now?

docs/scoping.md
```diff
@@ -0,0 +1,67 @@
...
+<div class='user-'>
+  //Internal HTML
+</div>
+```
+
+Scopes down into the Backbone Model with properties: `firstName`, `lastName`, and `hobby`.
+Notice this only allows scopping down.
+
+
+## Scoping Relatively (UNIX style)
```
bglusman: I think this should be actually scoping relatively or absolutely, / in front is absolute (like unix) and ./ or ../ are relative.  Should also clarify this is the only way to (temporarily) unscope/move back down the stack locally, where as the above style (class="attribute-") is kind of a short hand for data-scope="./attribute"

docs/scoping.md
devmanhinton: It is all over the docs, so I think yes :D

docs/scoping.md
```diff
@@ -0,0 +1,67 @@
...
+```html
+<div class='user-'>
+  <div data-scope='/'>
+    //Iternal HTML
+  </div>
+</div>
+```
+
+We scope down into user and then, via the data-scope property, scope back to the root level object
+(the literal array with propery `user`).
```
bglusman: I don't think you mean 'array' here, the root object with property user maybe?

devmanhinton: Ahh yes, thank you.

docs/scoping.md
```diff
@@ -0,0 +1,67 @@
...
+
+
+```html
+<div class='user-'>
+  //User scope
+  <div class='hobby-'>
+  //Hobby scope
+    <div class='..'>
+    //Back in User Scope
+      <div class='/user/hobby'>
```
bglusman: unless i'm badly confused neither this nor line 57 works, meant to be data-scope='../' on 57 and data-scope not class on 59 :-)

docs/scoping.md
bglusman: It is in a few other places... but that doesn't answer the question of whether it should be... they work the same, right?  why expose two to the world rather than one?  any difference/benefit?

docs/scoping.md
devmanhinton: whoops :p, yep. `data-scope`

docs/scoping.md
```diff
@@ -0,0 +1,69 @@
...
+Normal UNIX path shorthand apply: `..` to move back up a scope level, `/` to seperate scope levels,
+`.` for the current scope'.
+
+
+```html
+<div class='user-'>
+  //User scope
+  <div class='hobby-'>
+  //Hobby scope
+    <div data-scope='..'>
```
bglusman: Does this definitely work without the trailing /?  I'd always write this ../ but don't know if that's necessary or just me/allowed.  Kind of think it's clearer unix-wise, but no big deal, might just be me :-)

devmanhinton: I was not sure it worked with the slash :) I know it works without the slash since we use it that way in brief

devmanhinton:
closing in favor of #120

PR #119, opened 2013-11-08
duped in attributes

bglusman:
@devmanhinton

### DIFF
--- a/docs/enddash-syntax.md
+++ b/docs/enddash-syntax.md
@@ -18,12 +18,6 @@ template.bind({
 });
 ```
 
-## Attributes
-
-```html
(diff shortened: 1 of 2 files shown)


devmanhinton:
@bglusman Added deletion of other duplicate attribute documentation :D

devmanhinton:
If you approve can you comment & then I will merge

bglusman:
@devmanhinton oh I forgot... I think I only meant to remove the one you removed, maybe should leave the other one in syntax doc if we're leaving that around as an overview? your call, fine either way.

devmanhinton:
Yeah that sounds better to me. I'll push that up in another branch real quick

PR #130, opened 2013-11-08
Refactor partials and variable docs

devmanhinton:
@bglusman


bglusman:
incorporated changes into #132 with some tweaks/differences

PR #132, opened 2013-11-11
expand conditionals into two sections

bglusman:
@devmanhinton split into two sections and clarify a bit

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ Documentation
 
 [Conditionals](#conditionals)
   * [Ternary](#ternary)
-  * [EndDash Conditionals](#enddash-conditionals)
+  * [Visibility Conditionals](#visibility-conditionals)
 
 [Scoping](#scoping)
   * [What is scoping?](#what-is-scoping)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -330,34 +332,63 @@ different than length), as in the example below:
...
 that may be present, with or without a false condition:
 
 ```html
-<div class="user- #{availability ? available : unavailable}">
+<div class="user- #{availabile ? openings : unavailable}">
+</div>
+```
+
+or just
+
```
devmanhinton, line 346: Instead of 'or just'. 

'To add the a tag attribute only if the object attribute is true (or false) just
leave out the second clause.'

Or even cut this out. It seems like nice but extra syntax that a user might be able to figure out. What do you think?

bglusman, line 346: actually, can't do it if false, maybe with colon but without first condition?  dunno...I thought it was funny mentioning it but not showing an example, but I could go either way I guess

README.md
```diff
@@ -330,34 +332,63 @@ different than length), as in the example below:
...
+```html
+<div class="user- #{availabile ? openings }">
+</div>
+```
+
+The full ternary with false condition will add the `available` class when availability is
+truth, and the `unavailable` class when falsy.  The second example will only add the
+`available` class when availability is truthy.
+
```
devmanhinton, line 355: Hmmm I was confused when first reading this. It seems pretty explanatory now that I read it. However, that could easily because I have read it three times at this point :D

bglusman, line 355: Which part, the 'full ternary' phrasing? Open to edits :-)

README.md
devmanhinton, line 346: Yeah actually do not mention that, though one could always do `? "" : cake` to get the only when false behavior.

README.md
devmanhinton, line 355: A little :). Perhaps we can actually do away with this part? and then edit the 'or just' clause. (my other comment!)

bglusman, line 355: Yeah, dunno, feel free to edit and push to this branch... I'm poking at the custom template thing since it seems like if easily workable would be a big win, but can stop if you think it's silly...

README.md
```diff
@@ -330,34 +332,63 @@ different than length), as in the example below:
...
 ============
 
+## Ternary
+
 A ternary operator is available for presence handling via 'truthiness' for attributes
 that may be present, with or without a false condition:
 
 ```html
-<div class="user- #{availability ? available : unavailable}">
+<div class="user- #{availabile ? openings : unavailable}">
```
yaymukund, line 341: typo

bglusman:
closing in favor of #147

PR #134, opened 2013-11-11
Fix margins on footer + nav & main content

devmanhinton:
@adelbalso @FredericoAndrade 

This fixes a bit of style that was off. Basically, the nav menu items were flush against the left side of the page. To the extent that the start of the word looked like it was being cut off. 

This fixes that and fix a few other styles to make everything consistent.


bglusman:
think this is addressed, will have to rebase on new file name/locations if not

PR #135, opened 2013-11-11
bump bower & package.json to version 0.3.3

devmanhinton:
@bglusman  @yaymukund 

We forgot to update the bower.json in EndDash last time... :(


devmanhinton:
Clarification, I'm bumping to 0.3.3. But in the diff you will see the bower.json is jumping from 0.3.1 to 0.3.3

PR #136, opened 2013-11-11
Refactor tests for polymorphic iteration.

yaymukund:
@devmanhinton, @bglusman I just refactored some tests last week. I'm going to drop this now since it's bug week, but I think it's GTM. Please review.

### DIFF
--- a/test/polymorphic_looping.js
+++ b/test/polymorphic_looping.js
@@ -1,94 +1,58 @@
 require('./support/helper');
 
-var path = require("path"),
-    expect = require("expect.js"),
-    fs = require("fs"),
-    Backbone = require("backbone"),
-    generateTemplate = require("./support/generate_template");
+var expect = require('expect.js'),
+    fs = require('fs'),
+    Backbone = require('backbone'),
+    generateTemplate = require('./support/generate_template');
 
-describe("A polymporhic template", function() {
+describe('A collection template with polymorphic attributes', function() {
+  var markup =
+    '<ul data-each class="rolePolymorphic-">'+
+    '  <div class="#{tag} whenSuperhero-">I\'ll save you!</div>'+
+    '  <div class="#{tag} whenSupervillain-">I wont\'t.</div>'+
+    '  <div class="#{tag} default">I am going to run.</div>'+
+    '</ul>';
 
-  it("looping through a collection will display the correct item given the model type", function() {
-    var model = { things: [{ type: "awesome" }, { type: "cool" }] },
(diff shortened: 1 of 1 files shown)


test/polymorphic_looping.js
```diff
@@ -1,94 +1,60 @@
...
-    expect($(".things- li div:nth-child(2)").html()).to.be("cool");
+var expect = require('expect.js'),
+    fs = require('fs'),
+    Backbone = require('backbone'),
+    generateTemplate = require('./support/generate_template');
+
+describe('A collection template with polymorphic attributes', function() {
+  var markup =
+    '<ul class="characters-">'+
+    '  <li class="character- rolePolymorphic-">'+
```
devmanhinton: This is old style iteration 'characters-' 'character-' as opposed to new style 'data-each'

test/polymorphic_looping.js
```diff
@@ -1,94 +1,60 @@
...
-  });
-  it("will interpolate the values", function(){
-    expect($('.jobPolymorphic- div:nth-child(1)').html()).to.be('Scientists welcome you, signed by <div class="firstName-">Bill</div>');
-    expect($('.jobPolymorphic- div:nth-child(2)').html()).to.be('I am your teacher and I am <div class="kind-">tough</div>');
-    expect($('.jobPolymorphic- div:nth-child(3)').html()).to.be('Default here: <div class="name-">Joe</div>');
-    expect($('.jobPolymorphic- div:nth-child(4)').html()).to.be('Default here: <div class="name-">Rando</div>');
+    it('updates the case if the switch value changes', function() {
+      expect($('.superman.whenSuperhero-')).not.to.be.empty();
+      characters.findWhere({name: 'Superman'}).set('role', 'superhero');
+      expect($('.superman.whenSupervillain-')).not.to.be.empty();
```
devmanhinton, line 55: Did you mean to do

``` js
characters.findWhere({name: 'Superman'}).set('role', 'supervillain');
```

yaymukund:
Fixed both of those things. The `set('role', 'superhero')` typo didn't break the test, which is a bit worrisome.

devmanhinton:
@yaymukund Yeah... I was thinking that we should figure that out before merging maybe?

yaymukund:
@devmanhinton It breaks as expected with the updated syntax. I'm inclined to put the effort toward removing the deprecated syntax entirely (and maybe opening an issue just so we have a record of the possible bug.)

devmanhinton:
Ahhh I see. 

Agreed, if it only happens with the old syntax I agree.

PR #138, opened 2013-11-11
Add one line warning

devmanhinton:
@yaymukund 

See file changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Define your templates:
   </div>
 </script>
 ```
-
+WARNING: A template can only have one root element. In the above case, it is the div with class 'user'.
 
 Bind templates to models in your application code:
 ```javascript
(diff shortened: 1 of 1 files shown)


devmanhinton:
@bglusman

README.md
```diff
@@ -39,7 +39,7 @@ Define your templates:
   </div>
 </script>
 ```
-
+WARNING: A template can only have one root element. In the above case, it is the div with class 'user'.
```
yaymukund, line 42: How about `NOTE` or even `**Note**` ("**Note**") instead of `WARNING`?

bglusman, line 42: I think a warning is justified since an error will be raised otherwise, any specific reason you object to the word warning?  I think mentioning the error raised might make sense here also.  Formatting it with *\* instead of all caps might be an improvement.

PR #139, opened 2013-11-11
refactor partial docs

devmanhinton:
@yaymukund 

Makes partials much shorter and clearer

### DIFF
--- a/README.md
+++ b/README.md
@@ -598,14 +598,56 @@ Small, reusable, components of HTML can be templated in EndDash as partials.
 To use a partial, add `src='templateName'` as an attribute to an element with no children.
 
 ```html
-<div src='myPartial' data-replace></div>
+<script type='text/enddash' name='superheros'>
+  <img src="#{logo}" />
+  <div class='heros-'>
+    <div src='superhero-navigation' data-replace></div>
+  </div>
+</script>
 ```
+The partial will be passed the model in scope as its root element.
 
 The data-replace attribute tells EndDash to substitute the partial's root element for its partial.
 Without data-replace, EndDash will embed the root element beneath the partial's element and leave it.
 
-Load the partial you are referencing into EndDash before binding to the template referencing the partial.
+If elsewhere you define this partial as:
 
+```html
+<script type='text/enddash' name='superhero-navigation'>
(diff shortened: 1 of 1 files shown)


devmanhinton:
@bglusman

bglusman:
@devmanhinton think this is OK without any examples for partials?  why not have examples?

yaymukund:
Can you show a full example from loading a partial to using that partial?

And instead of `myPartial`, can you use a real example? Even something simple like:

``` html
<script type="text/enddash" name="superhero-navigation">
...
</script>

<div src="superhero-navigation"></div>
```

I do like the instinct to trim it down though. If your goal is to show how partials work, you don't need a collection with three things in it.

devmanhinton:
@bglusman I did just because templates & their syntax is loaded elsewhere.

@yaymukund I can add that. We document loading templates elsewhere (or rather there is the start of that code in a PR. After that merged I think I need to document boostrap().). I can add it here as well though?

yaymukund:
@devmanhinton I suggested it because people might not realize that loading partials is the same as loading templates...

devmanhinton:
@yaymukund Cool. No worries working on an edit to the affect right now.

devmanhinton:
@bglusman @yaymukund Good to check out again.

devmanhinton:
@bglusman

README.md
```diff
@@ -594,52 +594,59 @@ $('.content').html(boundTemplate.el);
...
-var person1 = new Backbone.Model({firstName: 'Tony', lastName: 'Stark', alias: 'IronMan'});
-var person2 = new Backbone.Model({firstName: 'James', lastName: 'Rhodes', alias: 'WarMachine'});
-var person3 = new Backbone.Model({firstName: 'Pepper', lastName: 'Potts', alias: 'none' });
-
-var people = new Backbone.Collection([person1, person2, person3]);
+```html
+<script type='text/enddash' name='superheroes'>
+  <img src="#{logo}" />
+  <div class='heroes-'>
+    <div src='superhero-navigation' data-replace></div>
```
yaymukund, line 604: Please use double quotes for HTML.

PR #140, opened 2013-11-11
Templates docs to readme

devmanhinton:
@yaymukund 

Tried to spend more time writing this so hopefully there are fewer small typos which I know bother you :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -533,13 +533,13 @@ Templates
 First, register a template in EndDash.
 
   ```js
-  EndDash.registerTemplate(templateName, template);
+  EndDash.registerTemplate('templateName', template);
   ```
 
 Next, bind to that template. To start, get the EndDash parsed template.
 
   ```js
-  var template = EndDash.getTemplate(templateName);
+  var template = EndDash.getTemplate('templateName');
   ```
 
 Then actually bind a model to that template.
@@ -551,7 +551,7 @@ Then actually bind a model to that template.
 The above two steps can be combined into one.
 
   ```js
-  var boundTemplate = EndDash.registerTemplate(templateName, myModel);
+  var boundTemplate = EndDash.registerTemplate('templateName', myModel);
   ```
 
 Next, display the HTML for this template on the page.
--- a/README.md
+++ b/README.md
@@ -532,23 +532,33 @@ Templates
 
 First, register a template in EndDash.
 
-  `EndDash.registerTemplate(templateName, template);`.
+  ```js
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -523,6 +525,41 @@ var views = {},
 EndDash.setCustomGetView(getViews);
 ```
 
+Templates
+=========
+
+## Registering & Binding
```
yaymukund, line 531: Can we de-abstract this? So something like:

``` js
// 1. Register a template in EndDash.
EndDash.registerTemplate('user', '<div class="user">#{name}</div>');

// 2. Fetch a bound template.
var user = new Backbone.Model({name: 'Mukund'}),
    template = EndDash.getTemplate('user', user);

// 3. Attach the bound template to your page.
$('.content', template.el);
```

Also, what about:

Right now, we're presenting the `getTemplate('foo').bind(foo)` as the default syntax and suggesting `getTemplate('foo', foo)` as an alternative. I feel like we should put our best foot forward and suggest the shorter `getTemplate('foo', foo)` syntax first.

devmanhinton, line 531: I would agree we want that syntax. I do, however, think it is valuable to explain it in this way first.

Show them how it can work with getTemplate then bind. And then show them a shortcut. 

I actually think it is good as is. Do you have anything against merging this? And then if you feel stylistically there is a better way, you put in the PR? I am pretty happy with the way it is now.

yaymukund, line 531: There were two suggestions in my previous comment. You only addressed the second one.

We should de-abstract examples. That is, show the actual markup string, use a template named `user` rather than `templateName`, generally present more concrete examples.

devmanhinton, line 531: In the future, can you indicate your reason along with your suggestion: ie., 'I think we should make examples less abstract `because I have found such examples helpful myself`. Otherwise, it seems a bit arbitrary.

In this case, I believe either is good. As some people learn better through abstraction and others through concrete examples :D. (for instance I often learn better through abstraction.) However, I also looked through backbone, ember, and angular. Backbone uses abstract and concrete while ember & angular both use only concrete examples.

For this reason,  I'll switch to concrete examples. Thank you for motivating this research and eventual conclusion :)

devmanhinton:
closing in favor of #143

README.md
yaymukund, line 531: Hm, I'll try to do this but remember you can always be like "Why are concrete examples better?" or "Why not do it this other way?"

PR #141, opened 2013-11-12
Remove unnecessary requires, clean up some code.

yaymukund:
@devmanhinton @bglusman We had far too many of these...


devmanhinton:
Sweet

PR #143, opened 2013-11-12
Add template docs

devmanhinton:
@yaymukund Concrete example docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,8 @@ Documentation
 
 [View Integration](#view-integration)
 
+[Templates](#templates)
+
 [Partials](#partials)
 
 [Debugger](#debugger)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -523,6 +525,49 @@ var views = {},
...
 
+Templates
+=========
+
+## Registering & Binding
+
+First, register a template in EndDash.
+
+  ```js
+  EndDash.registerTemplate('superpower',
```
yaymukund: Why have two interpolated variables instead of one? Shorter is better, unless you're trying to convey a complex point. With one interpolated varialbe, you can probably squeeze the markup into a single line..

devmanhinton: Sure. I made it longer to make it fun :D. 

But I see your point. Yes, the increased fun-ness does not make up for the length of it.

Will push up a shorter version.

devmanhinton:
@yaymukund Wooo. This is looking good. Also, add documentation of bootstrap :)

yaymukund:
Gonna merge this, since it's just docs. But I still think it's too verbose. Compare with something like:

``` html
<script type="text/enddash" name="superhero">
  <div class="superhero">
    Hi, I'm <span class="name-"></span>.
  </div>
</script>
```

And in your JS:

``` js
var hero = new Backbone.Model({ name: 'Superman' }),
    template = EndDash.getTemplate('superhero', hero);

$('.content').html(template.el);
```

Differences:
- End-to-end example first, save the alternate syntax for later.
- Shortest possible examples.
- Encourage the `<script>` template loading method. It's clearer separation between markup and JS, so I feel like it reads better in examples.

PR #146, opened 2013-11-12
Use existing backbone

### LINKED ISSUE #78: Defer to existing Backbone if one is present
@devmanhinton @bglusman 

When we initialize EndDash, we should defer to an existing backbone if one is already present.

In practice, I think it would look something like:

``` js
if (Backbone) {
  EndDash.Backbone = Backbone;
}

// ...
if (model instanceof EndDash.Backbone.Model) {
  ...
}
```

yaymukund:
@bglusman @devmanhinton 

Closes #78 

``` js
// Before
var Backbone = require('backbone');

// After:
var Backbone = require('end-dash').Backbone;
```

This branch includes backbone iff there isn't already a backbone on the page. Instead of checking if the object supports the `on`, `off`, `get`, `set` interface, we check if it's a Backbone object.

It also raises two questions:
- How can we clean up dependency management? Looks pretty ugly architecturally. Can we get away from using CommonJS on the client?
- Should we include Backbone if it doesn't already exist? If we're going to say Backbone is a dependency, maybe we should just error if it isn't present. This seems a lot clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@ EndDash
 
 EndDash is a bindings-aware client-side templating framework built on top of valid HTML.
 
-In its current release, EndDash relies on Backbone style objects.
-See [the dependency section](#dependencies) for further details.
+In its current release, EndDash relies on Backbone objects. See
+[the dependency section](#dependencies) for further details.
 
 ([Documentation](#documentation) is below)
 
@@ -26,9 +26,11 @@ grunt build # also aliased as `grunt`
 Include the library and dependencies.
 
 ```html
+<!-- Make sure you include these libraries before EndDash -->
 <script src="//ajax.googleapis.com/ajax/libs/jquery/1.10.2/jquery.min.js"></script>
 <script src="http://underscorejs.org/underscore.js"></script>
 <script src="http://backbonejs.org/backbone.js"></script>
+
 <script src="/end-dash/build/end-dash.js"></script>
 ```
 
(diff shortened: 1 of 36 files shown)


devmanhinton:
@yaymukund Do you mean to check in the backbone library?

devmanhinton:
This means we no longer support objects which define Backbone's interface. Which seems fine to me but should just be mentioned before we merge.

devmanhinton:
And this PR should include update to docs to reflect this difference as well :)

yaymukund:
@devmanhinton good catch. I've updated the docs accordingly, and rebased on top of latest master. Tests still pass.

devmanhinton:
@yaymukund Okay cool. I never heard. Did you mean to check in the Backbone library? This PR is almost 3000+ lines ...

yaymukund:
Yup, I did!

yaymukund:
@devmanhinton please review again. As per our discussion, I've [made the docs more specific about when we should load dependencies](https://github.com/Amicus/end-dash/pull/146/files#diff-04c6e90faac2675aa89e2176d2eec7d8R29). This documentation is hopefully only temporary given that we're going to require backbone explicitly in the near future.

I also rebased on most recent master, though that shouldn't affect anything.

devmanhinton:
@yaymukund O. I would have reviewed it again earlier. I thought you were writing that 'throw error if no backbone' now. I'll review it now

README.md
```diff
@@ -623,14 +625,9 @@ Misc
...
-a bound object's attributes into the DOM but will not update the DOM on model changes.
-
-
-
+In its current release, EndDash relies on Backbone's event system to update
+the DOM when a bound object changes. This means only templates bound to
+Backbone objects will live-update.
 
+If you use EndDash with regular JS objects, it'll still interpolate the
+attributes but won't update the template when the object's values change.
```
devmanhinton: odd that you are using attributes and then values both to refer an objects attributes. Change the second to 'attributes'?

Did you do use 'value' because it won't update if you change the models attributes or the model itself?

devmanhinton:
@yaymukund Can you fix that one line and then I will merge it :smile:

yaymukund:
@devmanhinton Fixed. You're right about the phrasing change— thanks.

PR #148, opened 2013-11-13
Include a js + ed => html example for polymorphic attributes.

yaymukund:
This fixes https://amicushq.atlassian.net/browse/AM-34.

Previously, the polymorphic attribute documentation included a template and HTML, but didn't include the models that were actually getting bound. 

Do you think this is clearer, @zonotope? 

Also, here's [a link to the compiled markdown!](https://github.com/Amicus/end-dash/blob/cc8349379cdc96aaa25cf10f45b953295f653421/README.md#polymorphic-attributes)

CC @devmanhinton @bglusman

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,70 +244,83 @@ var characters = new Backbone.Collection([
 
 ## Polymorphic attributes
 
-If your objects have an enum (or Enumerated type) field, you can specify handling based on which type it is.
-This is best explained with an example.
+If your objects have an enum field, you can branch based on its value.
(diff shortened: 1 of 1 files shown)


devmanhinton:
I like it! Will wait on Ben to comment since he is the requester! :)

yaymukund:
Er, fyi: I force-pushed a tiny (tiny!) fix.. changing "base case" to "default case." Still good to review/merge.

PR #153, opened 2013-11-14
Fixes AM-33 #resolved. Clarify simple looping and add a complete example

devmanhinton:
@zonotope Does this clarify looping as you requested?

cc @bglusman 

I tried to make the language clearer/ highlight 'data-each'. Additionally, I added a concrete example and its output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ Looping
 ## Simple Looping
 
 To reuse a set of DOM elements for each child model in a collection,
-add the ```data-each``` attribute to the parent of this set.
+add the `data-each` attribute to the parent of this set.
 
 ```html
   <div data-each>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -200,46 +200,45 @@ the attribute value on the input element (or 'true' if no value is defined).
...
+To reuse a set of DOM elements for each child model in a collection,
+add the ```data-each``` attribute to the parent of this set.
+
+```html
+  <div data-each>
+    <div class="firstName-"></div>
+  </div>
+```
+
+The object in scope at these elements, will be iterated through (using ```.each```).
```
yaymukund: For inline code snippets, you just need a single backtick like: ``.each`` (gets compiled to: `.each`)

devmanhinton: Ahhh Okay, thank you.

PR #155, opened 2013-11-14
Fix version on bower.json

yaymukund:
@devmanhinton D'oh!

Forgot to update bower.json to match package.json when updating to 0.3.4


devmanhinton:
@yaymukund lol. I should have realized before merging too :smile_cat:

PR #157, opened 2013-11-18
Javascript based is clearer than client side

bglusman:
language could probably use further fine tuning.  Final accepted version needs dupe to master also. @yaymukund @sethbannon @xcoderzach

### DIFF
--- a/_includes/README.md
+++ b/_includes/README.md
@@ -2,7 +2,7 @@
 Getting Started
 =======
 
-EndDash is a declarative two-way binding javascript based templating framework built on top of valid HTML.
+EndDash is a two-way binding javascript templating framework built on top of semantic HTML
 
 In its current release, EndDash relies on Backbone style objects.
 See [the dependency section](#dependencies) for further details.
--- a/_includes/README.md
+++ b/_includes/README.md
@@ -2,7 +2,7 @@
 Getting Started
 =======
 
-EndDash is a bindings-aware client-side templating framework built on top of valid HTML.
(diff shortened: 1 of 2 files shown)


_includes/README.md
```diff
@@ -2,7 +2,7 @@
 Getting Started
 =======
 
-EndDash is a bindings-aware client-side templating framework built on top of valid HTML.
+EndDash is a declarative two-way binding javascript based templating framework built on top of valid HTML.
```
yaymukund: How about "EndDash is a two-way binding javascript templating framework built on top of valid HTML"?

That's a lot of adjectives in a row. I sort of still prefer "bindings-aware Javascript templating framework." Those terms should all be familiar to anyone that is already researching this stuff. For the slogan, we want something catchy and conveys the gist. If it's confusing, we can expand later.

tobowers: we had at least 2 advanced JSers comment that they didn't know what "bindings-aware" means.

I like @yaymukund's rewrite there.  

alternatives:

EndDash is a javascript templating framework on top of valid HTML that allows two-way binding of Backbone objects to the DOM.

EndDash allows you to bind javascript objects to the DOM using semantic HTML.

bglusman: Yeah, mentioning Backbone's a probably a good idea in there... semantic HTML too.  Hmm... maybe top one with 'valid semantic HTML' ?

yaymukund: "valid, semantic" is redundant.

sethbannon:
Final language: "EndDash is a two-way binding javascript templating framework built on top of semantic HTML"

PR #168, opened 2013-11-18
syntactic not syntatic

devmanhinton:
@yaymukund  Fix simple spelling error


sethbannon:
This @amicus-jenkins guy is a real asshole

PR #170, opened 2013-11-18
Hack to fix weird syntax highlighting

devmanhinton:
@FredericoAndrade @bglusman 

This is a hack to get rid of weird syntax highlighting.

@yaymukund What do you think?


devmanhinton:
To clarify, we have HTML inside of script tags which github markdown reads as an error and adds the 'err' class.

http://www.enddash.com/#toc_25

FredericoAndrade:
to further clarify, the error gets thrown when there are closing html tags
inside a script element

Issue #172, opened 2013-11-20
Typo in a readme file


cf8:
There is, probably, a typo in the "Attribute Interpolation" section name here:
https://github.com/Amicus/end-dash/blob/e6935a5dcedbb499530243a2fcb504f6da3d54cc/README.md#attibute-interpolation

PR #173, opened 2013-11-20
fix for #172

jameselkins:
@bglusman tag

I am unsure whether I should change this in just the README.md, just the gh-pages branch, or both, so I'm opening this PR for a little bit of guidance. I heard there was a procedure that we were using so manual duplication didn't have to occur.


bglusman:
@jameselkins oh sorry missed this... yeah, changing in both is ideal but can't be done in one PR.  I'll fix in gh-pages.

PR #174, opened 2013-11-21
add component.json, and get rid of dev deps that are private

xcoderzach:
Starting a pull, don't merge it until you guys make xcoderzach/path an amicus repo.

Also, I got rid of test-things and awesome-docs as deps, since they aren't open source.

EDIT:

> they aren't open source.

yet O_o

### DIFF
--- a/component.json
+++ b/component.json
@@ -19,7 +19,6 @@
     "lib/page_helper.js",
     "lib/parser.js",
     "lib/reaction.js",
-    "lib/reactions",
     "lib/rules.js",
     "lib/template.js",
     "lib/template_store.js",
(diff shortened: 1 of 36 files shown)


xcoderzach:
So, I fixd all the cyclic dep errors, but didn't actually get rid of the cycles :-( .  I think that they're probably indicative of a larger problem.  I think that the real solution is to stop having the `EndDash` monolith object, and instead dependency inject things like `setCustomViewGetWhateverMyNameIsFunction`

tobowers:
@xcoderzach this looks ok to merge to me now.  Is that true?

PR #180, opened 2013-12-06
setTimeout on enddash variable setting for inputs

tobowers:
@yaymukund tag

There is a problem when using dynamically populated checkboxes/radio buttons /select fields where end dash will try to set the value based on a scope but the children haven't been parsed yet so it will error... for example:

``` html
<div class="present-">
  <select class="usersId-">
    <optgroup data-scope="/users" data-each>
      <option value="#{id}" class="name-"></option>
    </optgroup>
  </select>
</div>
```

When we traverse and get to the select.userId- we try to set val() on it, but the users have not yet populated.

This is not the "best" fix, but at least it allows the template to completely render before trying to set val().

### DIFF
--- a/lib/reactions/variable.js
+++ b/lib/reactions/variable.js
@@ -54,7 +54,12 @@ var VariableReaction = Reaction.extend({
   },
 
   init: function(next) {
-    this.setFrom(this.variableName);
+    if (this.isInput) {
+      var that = this;
+      setTimeout(function () { that.setFrom(that.variableName)}, 0);
+    } else {
+      this.setFrom(this.variableName);
+    }
     next();
   },
 
--- a/lib/reactions/variable.js
+++ b/lib/reactions/variable.js
@@ -54,13 +54,8 @@ var VariableReaction = Reaction.extend({
   },
 
   init: function(next) {
-    if (this.isInput) {
-      var that = this;
-      setTimeout(function () { that.setFrom(that.variableName)}, 0);
-    } else {
-      this.setFrom(this.variableName);
-    }
     next();
+    this.setFrom(this.variableName);
   },
 
   observe: function() {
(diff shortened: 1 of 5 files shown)


yaymukund:
Ugh, this is super ugly. If you push it like this, can you set a deadline to remove it? This shouldn't hang around production indefinitely.

tobowers:
@yaymukund I'm trying to think of what the "right" solution is and having trouble.  Ideas?

for select boxes the ideal is something like: "if I can't set the value because it's not an option, then wait for the option to appear and then try setting it" - but I'm not sure about that.

yaymukund:
IDK, probably one of:
- split parsing into two phases.
- instead of setting the value, attach an event that fires later that does this.

tobowers:
@xcoderzach @devmanhinton if you have any ideas :)

tobowers:
I'm closing this... but I'd still like thoughts if you have 'em

xcoderzach:
OH OH OH PICK MEMEMEMEMEMEMEEEEEEEEEEEEE!

do it after you call `next()`

next tells EndDash to parse the children, so anything you do after next happens after the children have been parsed.

xcoderzach:
35f8cbc makes your tests pass and doesn't need a setTimeout, w00t

xcoderzach:
actually, there isn't a test for this yet, one sec.

xcoderzach:
does it actually errror?  I wrote a test trying to replicate the issue and couldn't see it.

xcoderzach:
hurrrrrrr, my test was checking against "1" which is the first one, which is the default.

lol

xcoderzach:
On the bright side tho, I figured that out by using https://github.com/airportyh/testem.  Which, when paired with browserify is pretty easy to get going, and would probably be better than using JSDom.  I'll make a branch at some point.

xcoderzach:
WEWT, ready to merge.

Zach

tobowers:
@jameselkins @devmanhinton if you were waiting on this - here it is... we need to tag etc.

PR #197, opened 2014-04-26
Fix trigger dom issue

devmanhinton:
@tobowers @pbadger @xcoderzach 

This stops a `model.set` on the first setting of the DOM when binding the dome

a) Its not necessary (we are circularly setting in most cases or else setting `undefined` to `""`). This is because we have the model listen to the DOM and then we call `this.setFrom(variableName)`. This sets the `el` equal to the model value of the variable. This then triggers a DOM `change` (sometimes, if it doesn't that case is handled by setting `firstSetHappening` to false) which sets the property of the `el` back on the model d'oh! If the attribute on the model is `undefined`, we set that on the `el` but calling `el.val()` on a non-value input returns "". Thus we change `undefined` to `""`. 

In both cases, the listener for DOM changes does `model.set(key, that.el.val(), {validate: true})` Yikes! This calls Backbone.Model.validate (thus are validators, like on the registration page, when we don't want to). Solution is, use a boolean to know when its the first set and when its not. There are probably edge cases here, so please look for those.

CHA-:ear:s

### DIFF
--- a/lib/reactions/variable.js
+++ b/lib/reactions/variable.js
@@ -3,7 +3,7 @@ var Reaction = require("../reaction"),
 
 var VariableReaction = Reaction.extend({
 
-  setFrom: function(key, initialSet) {
+  setFrom: function(key) {
     var newVal = this.get(key),
         oldVal,
         shouldTriggerChange = false,
@@ -38,8 +38,10 @@ var VariableReaction = Reaction.extend({
           el.val(newVal);
         }
       }
-      if (shouldTriggerChange && !initialSet) {
+      if (shouldTriggerChange) {
         this.triggerDOMChange();
+      } else if (this.firstSetHappening) {
+        this.firstSetHappening = false
       }
     } else {
       oldVal = el.html();
@@ -55,7 +57,8 @@ var VariableReaction = Reaction.extend({
 
   init: function(next) {
     next();
-    this.setFrom(this.variableName, true);
+    this.firstSetHappening = true
+    this.setFrom(this.variableName);
   },
 
   observe: function() {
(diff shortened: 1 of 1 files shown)


tobowers:
this is good man.  Do you know when this was introduced? We aren't seeing that same problem in the version of enddash we're using.

devmanhinton:
@tobowers https://github.com/Amicus/end-dash/commit/7aecd9716c9171120b2ca5b567e60b0bf75048a2 || https://github.com/Amicus/end-dash/pull/180 for the PR

TLDR; :smile: 

From a) 

``` javascript
next()
this.setFrom(this.variableName);
```

To b)

``` javascript
this.setFrom(this.variableName);
next()
```

Binding a EndDash template uses `.bind` from the Template class on the passed in model. This traverses the DOM for EndDash template and calls
in Reaction.js

``` javascript
  this.init(function() {
    that.observe(function() {
      next(that.stack, true);
    });
    next(that.stack);
  });
```

Which breaks down to 'run init function of this reaction with a callback that will run this reactions `observe` function and then move on to its child (with the same pattern).' 

lib/reaction/variable.js init function was

```
this.setFrom(this.variableName);
next()
```

Meaning we set the variable before observing the DOM or setting up child elements (`variable.js#observe` is where it sets listeners on the DOM). That is finish both setting the DOM and start listening to the DOM to an element before moving on to its children.

The linked PR changes this order to allow for dynamic drop downs etc..

``` javascript
next()
this.setFrom(this.variableName);
```

This translates to, 'for each DOM element with a variable reaction, have its matching EndDash reaction listen to the DOM. Once you reach the root of the EndDash reaction tree, set the DOM from the bound models in reverse order (bottom up). This means with a form or dropdown (with dynamically populated children), you set all of the options before setting up the parent form. So that it will populate the default correctly. This is great! It just results in this bug. Where you have a circular set from the model, set back on the model. Which is really fine (we can discuss whether setting undefined to "" is okay) except for the fact we use `.set(key, value, {validate: true})` which basically prematurely triggers our validations

PR #198, opened 2014-05-01
Bump and cleanup fn

devmanhinton:
@yaymukund @tobowers @xcoderzach 

Indirectly added to EndDash's `cleanup` function in `Template.js#cleanup` (traverse EndDash structure and calls `reaction#stopObserving`.

We can optimize and do `model.off('change', null, this)`. Backbone.js iterates through an `_events` hash when you call `off` so its fast

Closing for now (spec writing time / improving)

### DIFF
--- a/lib/reaction.js
+++ b/lib/reaction.js
@@ -96,13 +96,14 @@ Reaction.prototype.change = function(property, callback, context) {
 
 Reaction.prototype.stopObserving = function() {
   this.el.off(".endDash" + this.cid);
-  this.stopListening(this._presenter);
+  this.eachModelOff()
 };
 
-Reaction.prototype.off = function() {
+Reaction.prototype.eachModelOff = function() {
+  var that = this;
   this.stopListening(this._presenter);
   _(this.relevantModels).each(function(relevantModel){
-    relevantModel.off(null, null, this);
+    relevantModel.off(null, null, that);
   })
 }
 
--- a/lib/reaction.js
+++ b/lib/reaction.js
@@ -28,6 +28,7 @@ Reaction.prototype.start = function(el, stack, next) {
   this.stack = stack.slice(0);
   this.model = _.last(this.stack);
   this._presenter = getPresenter(this.model);
+  this.relevantModels = [this._presenter]
 
   this.init(function() {
     that.observe(function() {
@@ -97,6 +98,13 @@ Reaction.prototype.stopObserving = function() {
   this.el.off(".endDash" + this.cid);
   this.stopListening(this._presenter);
 };
+
+Reaction.prototype.off = function() {
+  this.stopListening(this._presenter);
+  _(this.relevantModels).each(function(relevantModel){
+    relevantModel.off(null, null, this)
+  })
+}
 
 Reaction.parse = function() {};
 Reaction.afterParse = function() {};
--- a/lib/reaction.js
+++ b/lib/reaction.js
@@ -28,7 +28,7 @@ Reaction.prototype.start = function(el, stack, next) {
   this.stack = stack.slice(0);
   this.model = _.last(this.stack);
   this._presenter = getPresenter(this.model);
-  this.relevantModels = [this._presenter]
+  this.relevantModels = [this._presenter];
 
   this.init(function() {
     that.observe(function() {
@@ -102,7 +102,7 @@ Reaction.prototype.stopObserving = function() {
 Reaction.prototype.off = function() {
   this.stopListening(this._presenter);
   _(this.relevantModels).each(function(relevantModel){
-    relevantModel.off(null, null, this)
+    relevantModel.off(null, null, this);
   })
 }
 
--- a/test/stop_observing.js
+++ b/test/stop_observing.js
@@ -1,8 +1,10 @@
 require('./support/helper');
 
 var Model = require('../lib/end-dash').Backbone.Model,
+    Collection = require('../lib/end-dash').Backbone.Collection,
     expect = require("expect.js"),
-    generateTemplate = require("./support/generate_template");
+    generateTemplate = require("./support/generate_template"),
+    fs = require('fs');
 
 describe("When I clean up a template", function() {
   it("should remove listeners from the model on a variable", function() {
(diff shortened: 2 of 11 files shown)


lib/reaction.js
```diff
@@ -28,6 +28,7 @@ Reaction.prototype.start = function(el, stack, next) {
   this.stack = stack.slice(0);
   this.model = _.last(this.stack);
   this._presenter = getPresenter(this.model);
+  this.relevantModels = [this._presenter];
```
yaymukund, line 31: Why is relevantModels an array of a single presenter?

Relevant to what?

devmanhinton:
close and reopen to update diff

lib/reaction.js
devmanhinton, line 31: This says by default reactions listen to `this._presenter`

Relevant to the Reaction ('models that the Reactions are listening to')

yaymukund:
Doesn't this PR go against #5?

devmanhinton:
No it does not

yaymukund:
I do not understand your description of this branch:

> Indirectly added to EndDash's cleanup function in Template.js#cleanup (traverse EndDash structure and calls reaction#stopObserving.
> 
> We can optimize and do model.off('change', null, this). Backbone.js iterates through an _events hash when you call off so its fast

What is cleaning what up? Why are you storing models with every reaction? How does this affect `Template`? You mentioned indirect, but that doesn't actually explain the connection. I've read through the diff and don't see the traversal code that you refer to.

devmanhinton:
I added that description because the traversal code is not in the diff

> I've read through the diff and don't see the traversal code that you refer to.

If you go to Template.js#cleanup (as I wrote above) it is defined as so

``` javascript
Template.prototype.cleanup = function() {
    reaction.stopObserving();
    next(stack);
  });
};
```

As I described. 

> What is cleaning what up? 

Listeners (note the `model.off`)

> Why are you storing models with every reaction?

Necessary to turn of the listeners for that reaction. Please notice the `model.off(null, null, that)`
http://backbonejs.org/#Events-off

test/stop_observing.js
```diff
@@ -61,4 +63,38 @@ describe("When I clean up a template", function() {
...
+
+    it("Does not react", function(){
+      this.things.add(new Model({type: "awesome"}));
+      expect($('[data-each]').children().length).to.be(2);
+
+      this.things.reset();
+      expect($('[data-each]').children().length).to.be(2);
+    });
+
+    it("it does not error when you remove the HTML and then remove a model ", function(){
```
devmanhinton, line 88: @xcoderzach

lib/reaction.js
```diff
@@ -95,9 +96,17 @@ Reaction.prototype.change = function(property, callback, context) {
...
   this.el.off(".endDash" + this.cid);
-  this.stopListening(this._presenter);
+  this.eachModelOff()
 };
 
+Reaction.prototype.eachModelOff = function() {
+  var that = this;
+  this.stopListening(this._presenter);
+  _(this.relevantModels).each(function(relevantModel){
+    relevantModel.off(null, null, that);
```
yaymukund, line 106: (edit: nm, answered my own question :))

yaymukund:
Thanks for explaining in-person, @devmanhinton .

I would suggest something like:

``` javascript
// in lib/reaction.js#start
this.on('stopObserving', function() {
  this._presenter.off(null, null, this);
}, this);

// in lib/reactions/scope.js#execPathOnStack
this.on('stopObserving', function() {
  prev.off(null, null, this);
}, this);

// in lib/reaction.js#stopObserving
this.trigger('stopObserving');
```

devmanhinton:
@yaymukund Why is that better?

Also, looked into hooking into JQuery

Can't do unless we include JQuery UI (which contains a `remove` event being trigger)

We can do it without JQuery using 

``` javascript
$("body").bind('DOMNodeRemoved', function(evt){
if (evt.target === that.template.el[0] // Get DOM not JQuery Object
  that.template.cleanup()
})
```

Doesn't work in ie8 :fist: So need to find an alternative

devmanhinton:
FYI ^ is fine even if we have > 1 copy of the same template on a page because `JQuery#clone` returns a new object

``` javascript
//From Template.js#Template
this.el = this.template = this.markup.clone(); 
```

yaymukund:
What do you think is best?

devmanhinton:
Without, it's an unnecessary level of indirection to listen for an event you trigger on yourself

Usually events are used when you have a third shared object and need to communicate through it. For instance, in brief. A model is shared between two views. View A has to take some action after View B after some DOM event has happened. View A and B have no reference to each other. Thus on DOM event, View A triggers an event on the model to indicate to B it is time for it to act on the model

Here the Template is traversing the EndDash structure and already has a reference to the Reaction. So it is clearer to simply call a function to stopObserving

yaymukund:
I just think `.relevantModels` is a poorly defined array and was trying to get around that.

Relevant to what? You said it's "relevant to the reaction" but everything on a reaction is relevant to the reaction by definition. If you rename it to `this.models`, then how does it relate to `this.model`? Can you think of any other uses for that array?

devmanhinton:
> Can you think of any other uses for that array?

No, its for this purpose.... which is fine...

> everything on a reaction is relevant to the reaction by definition

I see your point. It seems like your gripe is with naming and not functionality. `this.modelsListeningTo`?

yaymukund:
You don't see my point. My point is not about the naming.

devmanhinton:
Then clarify your point, what is it?

yaymukund:
Hey, I'm gonna let @tobowers finish up this review. I've already sunk a lot of time into this today and I feel like this isn't going anywhere.

devmanhinton:
@yaymukund Okay, cool, it'd be helpful if you could document what your gripe with the methodology presented in this PR is. No worries if you are too busy with your availableMailRecipients specs

devmanhinton:
The other option moving forward is to eliminate all this code and then change EndDash reactions to always use `listenTo` and never use `on`. Then all of this is irrelevant and we can simply call `this.stopListening` where `this` is a reaction object

devmanhinton:
Closing in favor of last comment. That is a much cleaner implementation and solves your concerns (at least the one I understood regarding presenters). That was actually my first intuition .... >_< 

Silly Brain, thinking is your thing, why fail at it sometimes ??

PR #199, opened 2014-05-02
Fix cleanup

devmanhinton:
This PR changes all external listeners in EndDash to use `listenTo` so that cleanup is as easy as `this.el.off(); this.stopListening(this._presenter);`. (exception is scope.js which utilizes a more complex schema). This fixes the bug that right now `Template.js#cleanup` does not actually cleanup all the listeners :smile: 

@yaymukund @tobowers @xcoderzach

### DIFF
--- a/lib/reaction.js
+++ b/lib/reaction.js
@@ -95,7 +95,7 @@ Reaction.prototype.change = function(property, callback, context) {
 
 Reaction.prototype.stopObserving = function() {
   this.el.off(".endDash" + this.cid);
-  this.stopListening(this._presenter);
+  this.stopListening();
 };
 
 Reaction.parse = function() {};
--- a/lib/reactions/debugger.js
+++ b/lib/reactions/debugger.js
@@ -19,7 +19,7 @@ var Reaction = require('../reaction');
 
 var DebuggerReaction = Reaction.extend({
   init: function() {
-    this.model.on('sync', this.startDebugger.bind(this));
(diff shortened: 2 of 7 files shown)


yaymukund:
Why is `this.el.off(); this.stopListening(this._presenter);` a better approach than [the code that I had pasted in the last PR](https://github.com/Amicus/end-dash/pull/198#issuecomment-41948989)?

devmanhinton:
@yaymukund :smile: I've coded and spec this method. If you have a _reason_ another way is better then please explain why

yaymukund:
In this branch, there are three ways to tear down listeners (`stopListening`, `trigger('stopListening')`, and `this.el.off`). In the other way, there's only one. (`this.on('stopObserving', ...)`)

tobowers:
the listenTo approach is nice because of GC and having a clear way to catch all listeners present and future... but... yeah, 3 ways to stop is a little weird... would doing the trigger approach but using stopListening help?

This is a teardown right, don't you only need the this.stopListening()  (without an object) once and then we don't care about the prev?

As in:

``` js
  stopObserving: function(){
    this.el.off(".endDash" + this.cid);
    this.stopListening();
  },
```

`this` in the listener for stop listening to `prev` is the same scope, I think?  (I could be wrong)

devmanhinton:
@tobowers o, lol. Of course. Thats cleanest.

I'll change that one `stopObserving`

yaymukund:
I'll just let @tobowers finish up this PR then.

PR #203, opened 2014-05-13
Remove listeners from nested looping reactions

devmanhinton:
@tobowers @zonotope 

Blocking postcards refactor

Nested looping reactions were not being cleaned properly

### DIFF
--- a/test/stop_observing.js
+++ b/test/stop_observing.js
@@ -159,4 +159,29 @@ describe("When I clean up a template", function() {
       });
     });
   });
+
+  describe("A template with nested looping", function() {
+    describe("which has data-scope attributes", function() {
+      before(function() {
+        this.dude = new Model({name: 'dude'});
+        this.item = new Model({name: 'item'});
+        this.nestedCollection = new Collection([this.dude, this.item]);
+        this.nestedCollectionModel = new Model({nested: this.nestedCollection});
+        this.collection = new Collection(this.nestedCollectionModel);
+        this.template = generateTemplate(this.collection,'<div id="head" data-each>\
+            <div class="nested-" data-each>\
+              <span class="name-"></span>\
+            </div>\
+           </div>'
+        );
+        this.template.cleanup();
+        $('body').html('') // Remove data properties from DOM
+      });
+
+      it("should remove listeners from the nested collection", function() {
+        this.nestedCollection.pop() // Note, this was erroring beforehand
+                                    // Not expecting an error because Mochajs w/ virtual DOM has a bug :(
+      });
+    });
+  });
 });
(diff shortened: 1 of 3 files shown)


devmanhinton:
@yaymukund @pbadger To keep you in the loop (since relevant for postcards refactor)

test/stop_observing.js
```diff
@@ -159,4 +159,29 @@ describe("When I clean up a template", function() {
...
+              <span class="name-"></span>\
+            </div>\
+           </div>'
+        );
+        this.template.cleanup();
+        $('body').html('') // Remove data properties from DOM
+      });
+
+      it("should remove listeners from the nested collection", function() {
+        this.nestedCollection.pop() // Note, this was erroring beforehand
```
tobowers: is there anyway we can actually test the behavior here?

devmanhinton: I can do some hackery to save the `el` that is removed via `$('body').html('')`. Really, what I want to test is that when I pop the model it does not try to update the DOM (which has had all its data removed) which will error

I ran this with and without the fix in this PR to make sure it tested what I thought

I think this is good as is but am open to adding in the saving of the el (I am really testing a bug/error does not happen)

tobowers: how about just using the expect.js throwException then? `expect(fn2).to.not.throwException();` https://github.com/LearnBoost/expect.js/ so at least it's clear

test/stop_observing.js
```diff
@@ -159,4 +159,29 @@ describe("When I clean up a template", function() {
...
+                                    // Not expecting an error because Mochajs w/ virtual DOM has a bug :(
```
devmanhinton: @tobowers Because of this comment, I tried that and it errors for me. It seems like the virtual DOM framework is overwriting that expect.js helper. It kept falsely erroring for me. That was indeed my first intention

tobowers: I see.  Would you explain the bug in the comment?  Maybe a link out to the bug description.

test/stop_observing.js
tobowers: I changed the comment a little bit with a blank to fill in :)

devmanhinton: Thank you Topper :P I actually don't understand what the error is, I figured it was not worth my time to delve into the error of another library since I can test without it. Can we agree just to clarify why I am testing as such and not figure out exactly what the helper is erroring?

tobowers: don't have to know the line of code or whatever, but we do need to know what's happening from a high level

devmanhinton:
@tobowers Fixed, add back in the the exception syntax, turns out I was calling it incorrectly -- now looks good to go :)

